PR #13, opened 2022-02-17
jackmeyer/cf-559-browser-prompt

meyerjrr:


pkg/browsers/detect.go
```diff
@@ -242,7 +242,7 @@ func HandleBrowserWizard(ctx *cli.Context) error {
 	if strings.Contains(strings.ToLower(browserName), "chrome") {
 		fmt.Fprintf(os.Stderr, "ℹ️  Granted has detected that your default browser is a Chromium based browser: %s\n", GetBrowserName(browserName))
 
-		label := "Do you want to select a different browser as the default?"
+		label := "Do you want to keep this as your default browser?"
```
chrnorm, line 245: I'm wondering if this might confuse users and give them the impression that they're changing their system's default browser, rather than the browser that Granted uses.

Thinking a bit more about this I think a good CLI output might be:
```
➜ assume
Granted allows you to access multiple cloud accounts in your web browser (https://granted.dev/multi-account).
It works best with Firefox but we also support Chrome, Brave, and Edge.

ℹ️ We've detected that your default browser is Chrome.

Use this browser with Granted? Y/n
```


```suggestion
		label := "Use this browser with Granted?"
```

chrnorm (changes requested):
thanks for bearing with me on these minor changes @meyerjrr - I've got a small further UX improvement to suggest around the wording of the initial prompt.

PR #29, opened 2022-02-22
Tidy up

JoshuaWilkes:


JoshuaWilkes:
Note we will need to update the build workflows to set the granted cli path correctly

meyerjrr (approved):

PR #30, opened 2022-02-22
Jackmeyer/cf-555 add   service flag which allows the user

meyerjrr:
https://linear.app/exponent/issue/CF-563/running-assume-console-without-arguments-should-open-the-console-for
- When a role is actively set running `assume -ar` will open up the console using the active role
https://linear.app/exponent/issue/CF-555/add-service-flag-which-allows-the-user-to-open-the-console-on-a-given
- Opens the console at the specified service with the -service flag
https://linear.app/exponent/issue/CF-554/add-region-flag-which-allows-the-user-to-specify-a-region-to-open-the
- Opens the console with the specified region 
Extra:
Updated browser prompts:
- granted browser: Now returns the current active default browser
- granted browser -set: Runs the change browser wizard

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -59,7 +59,7 @@ func assumeCommand(c *cli.Context) error {
 		}
 	}
 	//set the sesh creds using the active role if we have one and the flag is set
-	if c.Bool("active-role") && os.Getenv("AWS_ROLE_PROFILE") != "" {
+	if c.Bool("active-role") && os.Getenv("GRANTED_AWS_ROLE_PROFILE") != "" {
 		//try opening using the active role
 		fmt.Fprintf(os.Stderr, "Attempting to open using active role...\n")
 
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -105,8 +105,9 @@ func assumeCommand(c *cli.Context) error {
 	if openBrower && isIamWithoutAssumedRole {
 		fmt.Fprintf(os.Stderr, "Cannot open a browser session for profile: %s because it does not assume a role", profile.Name)
 	} else if openBrower {
+		service := c.String("service")
 		fmt.Fprintf(os.Stderr, "Opening a console for %s in your browser...", profile.Name)
-		return browsers.LaunchConsoleSession(sess, labels)
+		return browsers.LaunchConsoleSession(sess, labels, service)
 	} else {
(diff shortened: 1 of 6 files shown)


scripts/assume
```diff
@@ -27,6 +27,7 @@ if [ "$GRANTED_FLAG" = "GrantedAssume" ]; then
   unset AWS_ACCESS_KEY_ID
   unset AWS_SECRET_ACCESS_KEY
   unset AWS_SESSION_TOKEN
+  unset AWS_ROLE_PROFILE
```
JoshuaWilkes, line 30:
```suggestion
  unset GRANTED_AWS_ROLE_PROFILE
```

scripts/assume
```diff
@@ -39,6 +40,9 @@ if [ "$GRANTED_FLAG" = "GrantedAssume" ]; then
   if [ ! "${GRANTED_3}" = "None" ]; then
     export AWS_SESSION_TOKEN=${GRANTED_3}
   fi
+  if [ ! "${GRANTED_4}" = "None" ]; then
+    export AWS_ROLE_PROFILE=${GRANTED_4}
```
JoshuaWilkes, line 44:
```suggestion
    export GRANTED_AWS_ROLE_PROFILE=${GRANTED_4}
```

pkg/assume/assume.go
```diff
@@ -58,8 +58,18 @@ func assumeCommand(c *cli.Context) error {
 			}()
 		}
 	}
+	//set the sesh creds using the active role if we have one and the flag is set
+	if c.Bool("active-role") && os.Getenv("AWS_ROLE_PROFILE") != "" {
```
JoshuaWilkes, line 62:
```suggestion
	if c.Bool("active-role") && os.Getenv("GRANTED_AWS_ROLE_PROFILE") != "" {
```

pkg/assume/assume.go
```diff
@@ -58,8 +58,18 @@ func assumeCommand(c *cli.Context) error {
...
+		//try opening using the active role
+		fmt.Fprintf(os.Stderr, "Attempting to open using active role...\n")
 
-	if profile == nil {
+		profileName := os.Getenv("AWS_ROLE_PROFILE")
```
JoshuaWilkes, line 66:
```suggestion
		profileName := os.Getenv("GRANTED_AWS_ROLE_PROFILE")
```

JoshuaWilkes (changes requested):
added comments to rename the env var, otherwise good to go

PR #34, opened 2022-02-23
Jackmeyer/cf-589 include the region name in the firefox

meyerjrr:


pkg/browsers/console.go
```diff
@@ -124,8 +124,15 @@ func OpenWithChromiumProfile(url string, labels RoleLabels, selectedBrowser Brow
 		// A possible approach would be to open chrome in a headless way first then open it fully after setting the name
 
 		userDataPath := path.Join(grantedFolder, "chromium-profiles", fmt.Sprintf("%v", selectedBrowser))
+
+		//check the file path
+		_, err = os.Stat(chromePath + "/" + labels.MakeExternalProfileTitle())
```
JoshuaWilkes, line 129: @meyerjrr What is this check for? seems like this would fail every time because the chrome path is the application path not the profile path, and the profiles are created automatically by the chrome application not by us

JoshuaWilkes (approved):

PR #35, opened 2022-02-23
Josh/cf 585 granted gives invalid grant if the

JoshuaWilkes:


scripts/assume.fish
```diff
@@ -34,6 +34,9 @@ else if test "$GRANTED_FLAG" = "GrantedAssume"
   if test "$GRANTED_4" != "None"
```
meyerjrr, line 34: On line 23 of this script you might need to add `set -e AWS_REGION`

PR #36, opened 2022-02-23
Josh/cf 592 granted browser should detect and show

JoshuaWilkes:


pkg/browsers/windows.go
```diff
@@ -0,0 +1,35 @@
+package browsers
```
meyerjrr, line 1: Nitpick but should we put these structs into the osx file rather than the windows onw

JoshuaWilkes: Good pickup, I hadn't checked where these belonged, assumed it was windows

PR #37, opened 2022-02-23
Fish autocompletions to update, dev build support

jordiup:
image
image
image

### DIFF
--- a/pkg/granted/completion.go
+++ b/pkg/granted/completion.go
@@ -6,6 +6,7 @@ import (
 	"os/user"
 
 	"github.com/common-fate/granted/internal/build"
+	"github.com/common-fate/granted/pkg/assume"
 	"github.com/fatih/color"
 	"github.com/urfave/cli/v2"
 )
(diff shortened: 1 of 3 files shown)


pkg/granted/completion.go
```diff
@@ -39,8 +46,13 @@ var CompletionCommand = cli.Command{
...
 			f, err := os.Create(executableDir)
 
+			// If that fails, try to open an existing file
 			if err != nil {
-				fmt.Fprintln(os.Stderr, "Something went wrong when saving fish autocompletions: "+err.Error())
+
+				f, err = os.Open(executableDir)
+				if err != nil {
+					fmt.Fprintln(os.Stderr, "Something went wrong when saving fish autocompletions: "+err.Error())
+				}
```
jordiup, line 55: These lines are important, ensures the file is overwritten if it already exists

JoshuaWilkes: @jordiup I think you can use a builtin method in Go to open/create then write to a file

os.WriteFile(executableDir, []byte(output), 0600) 

This should handle everything for you and you wont need to create or open yourself

JoshuaWilkes (approved):

PR #39, opened 2022-02-23
Jordi/cf 594 add ux prompt for adding roles

jordiup:


pkg/assume/assume.go
```diff
@@ -60,6 +61,12 @@ func AssumeCommand(c *cli.Context) error {
 		var p string
 		err = testable.AskOne(&in, &p, withStdio)
 		if err != nil {
+			if strings.Contains(err.Error(), "please provide options") {
```
JoshuaWilkes, line 64: I think this check should go before the prompt.

We can check if the list of profiles is empty instead

JoshuaWilkes (approved):

PR #45, opened 2022-02-24
Support modifier flags such as '-c' being placed after the profile name

JoshuaWilkes:


pkg/assume/assume.go
```diff
@@ -23,6 +28,8 @@ func AssumeCommand(c *cli.Context) error {
 	}
 
 	var profile *cfaws.CFSharedConfig
+	ca := c.Args().Slice()
+	_ = ca
```
chrnorm, line 32: is this needed?

pkg/urfav_overrides/flags.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/urfave/cli/v2"
+)
+
+type Flags struct {
+	*flag.FlagSet
+	urFavFlags []cli.Flag
+}
+
+func New(name string, flags []cli.Flag, c *cli.Context) (*Flags, error) {
```
chrnorm, line 19: If you could put a quick description as to why we need the override that would be fantastic - just to avoid causing confusion as to why this is here.

chrnorm:
Looks great @JoshuaWilkes and very keen to get this merged. I've left a couple of review comments and also there are some conflicting files with `main` on this branch!

pkg/assume/assume.go
meyerjrr: Ill take a look!

PR #47, opened 2022-02-24
Fix Mac Chrome Path

hhwagner:
i think that path is wrong


chrnorm:
🙌 thanks for picking this up @hhwagner.

PR #61, opened 2022-02-25
Reg lookup windows default browser

Aiden-Ziegelaar:
Added ability to lookup default browser on Windows, uses the registry so had to use build tags/extensions to get cross platform compiles, tested on Windows and Mac M1


Aiden-Ziegelaar:
Just realised you're providing binaries for windows arm64 and i386, I'll update the build tagging in the morning

Aiden-Ziegelaar:
Build tagging updated, should work for the other architectures however I don't have a device to test.

meyerjrr:
Thanks for the contribution @Aiden-Ziegelaar! 
Just updated the file names and reopened the PR here #65

PR #80, opened 2022-03-02
Fix comment typo

QuinnyPig:
Someone might get hurt with this if they copy and paste blindly.


chrnorm (approved):
The alternative approach here would be to introduce a new `commamds` folder (like `commands`, but better)?

Thanks for the fix @QuinnyPig!

Issue #78, opened 2022-03-02
Feature Request: Support use on "headless" nodes


QuinnyPig:
Granted bombs out on initial run on Linux if xdg-settings isn't installed. A usage model where it spits out the URL for the console to STDOUT is normally how I've seen other SSO-driven tools handle this; is there a plan to support that approach?

meyerjrr:
Hey @QuinnyPig! 
Thanks for the suggestion, definitely worth having the url passed to the user approach. I just want to get to the bottom of the error you received when trying in your headless node.
1. Were you able to retrieve the credentials using Granted -> errored out attempting to open the browser?
2. Were you unable to get the creds due to not being able to run a sso console device flow? 
I think the error message you received will indicate which of these occurred

QuinnyPig:
Just noticed it was version 0.1.3, if that's relevant; installed via following the instructions in the quickstart.

`assume` bombs out with:

 ```
Granted works best with Firefox but also supports Chrome, Brave, and Edge (https://granted.dev/browsers).
exec: "xdg-settings": executable file not found in $PATH

```

meyerjrr:
Ahh it's falling over when attempting to find your default browser on linux. Thanks for the added info! 
Should probably silently error here and let the user specify their browsers path

QuinnyPig:
I’d also request that STDOUT be one of the browser options.

PR #82, opened 2022-03-03
Fallback url link if browser opening fails

meyerjrr:
Fallback url given to users if opening browser fails
-u flag for returning link rather than browser opening

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -99,6 +99,33 @@ func AssumeCommand(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+
+	if assumeFlags.Bool("url") {
+		//dont want to open the browser just return the link
+		labels := browsers.RoleLabels{Profile: profile.Name}
+
+		var creds aws.Credentials
+
+		creds, err = profile.AssumeConsole(c.Context)
+		if err != nil {
+			return err
+		}
+
+		service := assumeFlags.String("service")
+		if assumeFlags.String("region") != "" {
+			region = assumeFlags.String("region")
+		}
+
+		labels.Region = region
+		labels.Service = service
+		err, url := browsers.MakeUrl(browsers.SessionFromCredentials(creds), labels, service, region)
+		if err != nil {
+			return err
+		}
+
+		browsers.ManuallyOpenURL(url)
+		return nil
+	}
 	openBrower := assumeFlags.Bool("console") || assumeFlags.Bool("active-role")
 	if openBrower {
 		// these are just labels for the tabs so we may need to updates these for the sso role context
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -120,7 +120,7 @@ func AssumeCommand(c *cli.Context) error {
 
 		labels.Region = region
 		labels.Service = service
-		err, url := browsers.MakeUrl(browsers.SessionFromCredentials(creds), labels, service, region)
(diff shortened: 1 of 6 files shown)


JoshuaWilkes (commented):
@meyerjrr Looks good, though with this there is no way to get the firefox container url.

I think this should render the URL for the default browser you have configured, or possibly allow the browser to be specified?

DE = if firefox is my default/configured browser, the -url flags outputs the firefox container url

meyerjrr:
#78

pkg/browsers/console.go
```diff
@@ -126,7 +143,7 @@ func OpenWithFirefoxContainer(urlString string, labels RoleLabels) error {
 		}
 	}
 
-	tabURL := fmt.Sprintf("ext+granted-containers:name=%s&url=%s", labels.MakeExternalFirefoxTitle(), url.QueryEscape(urlString))
+	tabURL := fmt.Sprintf("ext+granted-containers:name=%s&url=%s", labels.Profile+labels.MakeExternalProfileTitle(), url.QueryEscape(urlString))
```
meyerjrr, line 146: Small change here to update the profile name to exclude the region but still have a unique id

pkg/assume/assume.go
```diff
@@ -99,6 +101,48 @@ func AssumeCommand(c *cli.Context) error {
...
+		cfg, _ := config.Load()
+		if cfg == nil {
+			return browser.OpenURL(url)
+		}
+		if cfg.DefaultBrowser == browsers.FirefoxKey {
+			url = browsers.MakeFirefoxContainerURL(url, labels)
+			if err != nil {
+				return err
+			}
+			browsers.ManuallyOpenURL(url)
```
JoshuaWilkes, line 137: can we replace this with the stdout print, I think we can get away with not printing a friendly message here because you get this result by specifying the -url flag, so you expect to get a url
```suggestion
			fmt.Print(url)
```

pkg/assume/assume.go
```diff
@@ -99,6 +101,48 @@ func AssumeCommand(c *cli.Context) error {
...
+
+		} else {
+			browsers.ManuallyOpenURL(url)
```
JoshuaWilkes, line 140:
```suggestion
			fmt.Print(url)
```

JoshuaWilkes (changes requested):

pkg/browsers/console.go
```diff
@@ -116,13 +132,16 @@ func OpenWithFirefoxContainer(urlString string, labels RoleLabels) error {
...
```
JoshuaWilkes, line 135: @meyerjrr can you update the MakeExternalFirefoxTitle() function to do this so its reusable?

Also I think you should replace this taburl with a call to MakeFirefoxContainerURL() to avoid duplicating

JoshuaWilkes, line 135:
```suggestion
	tabURL := MakeFirefoxContainerURL(ursString, labels)
```

pkg/browsers/console.go
```diff
@@ -181,10 +201,10 @@ const (
 	BrowserDefault
 )
 
-func LaunchConsoleSession(sess Session, labels RoleLabels, service string, region string) error {
+func MakeUrl(sess Session, labels RoleLabels, service string, region string) (error, string) {
```
JoshuaWilkes, line 204: Nitpick, can this be (string, error) as is convention for go functions returning errors, the error is the last value

Issue #81, opened 2022-03-02
Feature Request: command to unset environment variables


QuinnyPig:
If I `assume shitposting`, granted does what you would expect; sets the relevant environment variables to reflect that session.

Once I'm done with my various shitposting-related tasks, I either have to manually (or via a crappy shell script) unset those variables (lest I inadvertently do some production work in a shitposting environment) or else close the shell. 

A better approach would be to have a command that explicitly unsets those environment variables and returns my `env` to its pristine state. I nominate `assume not` as the command name.

QuinnyPig:
Come to think of it, an `assume logout` that revokes the SSO oauth token is also probably needed as well...

Issue #79, opened 2022-03-02
Feature Request: Custom environment variables


QuinnyPig:
Being able to set (for example) the AWS_PROFILE environment variable rather than having to teach an ecosystem of tools to respect GRANTED_AWS_PROFILE / write a shim that does it for me would be swell!

PR #83, opened 2022-03-03
command to unset environment variables

meyerjrr:
Added `assume revoke`

- Plan in the future is to move all commands surrounding configuration and cleanup to be moved to the Granted binary.


meyerjrr:
#83

JoshuaWilkes:
Fixes #79

chrnorm:
This should also fix #81!

JoshuaWilkes (approved):

PR #97, opened 2022-03-08
problems with assuming roles

### LINKED ISSUE #69: Problems with Assuming roles with MFA
Hello, when i try to assume a role with mfa the following Error occur:
` assume role with MFA enabled, but AssumeRoleTokenProvider session option not set.`

I think its because the SDK doesn’t automatically set the AssumeRoleTokenProvider with a default value. This is because of the risk of halting an application unexpectedly while the token provider waits for a nonexistent user to provide a value due to a configuration change. You must set this value to use MFA roles with the SDK.

### LINKED ISSUE #87: Handle global services automatically
For some services like Route53, if you specify that to open the console to, it tries to go to the regional version, and that one especially is global. Others I'm aware of are chime, iam and s3, though there are likely more.

meyerjrr:
Fixes #69 Fix problems assuming roles with MFA
Update unset command text to be more accurate

Fixes #87 Added global services list to check for url creation


pkg/browsers/console.go
```diff
@@ -50,6 +50,12 @@ var ServiceMap = map[string]string{
 	"iam":            "iamv2",
 }
 
+var globalServiceMap = map[string]string{
+	"iam":     "iam",
+	"route53": "route53",
+	"r53":     "r53",
+}
```
chrnorm, line 57: If we don't need to lookup anything this might be better as `map[string]bool` as it was a bit confusing whether these are meant to indicate particular URLs in AWS for global services. The `r53` doesn't actually point to `r53.amazonaws.com` as an example.

```suggestion
var globalServiceMap = map[string]bool{
	"iam":     true,
	"route53": true,
	"r53":     true,
}
```

pkg/browsers/console.go
```diff
@@ -304,9 +310,12 @@ func makeDestinationURL(service string, region string) (string, error) {
...
 
-	//NOTE here: excluding iam here and possibly others as the region isnt in the uri of the webpage on the console
-	if region != "" || serv != "iam" {
+	//excluding region here if the service is apart of the global service list
+	//uncomplete list of global services
+	notGlobal := globalServiceMap[service] == ""
+	if notGlobal && region != "" {
 		dest = dest + "?region=" + region
+
 	}
```
chrnorm, line 319: A more 'Go-like' way to do this is using the second return value from a map lookup to check whether it exists.
```
val, ok := dict["foo"]
```

```suggestion
	_, global := globalServiceMap[service]
	hasRegion := region != ""
	
	if !global && hasRegion {
		dest = dest + "?region=" + region
	}
```

pkg/browsers/console.go
```diff
@@ -284,9 +290,13 @@ func makeDestinationURL(service string, region string) (string, error) {
 
 	dest := prefix + serv + "/home"
 
-	//NOTE here: excluding iam here and possibly others as the region isnt in the uri of the webpage on the console
-	if region != "" || serv != "iam" {
+	//excluding region here if the service is apart of the global service list
+	//uncomplete list of global services
```
JoshuaWilkes, line 294:
```suggestion
	//incomplete list of global services
```

pkg/browsers/console.go
```diff
@@ -284,9 +290,13 @@ func makeDestinationURL(service string, region string) (string, error) {
...
+	_, notGlobal := globalServiceMap[service]
```
JoshuaWilkes, line 295: The map lookup will return true if the item is in the map, so _, global := globalServiceMap["iam"], global would be true

```suggestion
	_, global := globalServiceMap[service]
```

pkg/browsers/console.go
```diff
@@ -284,9 +290,13 @@ func makeDestinationURL(service string, region string) (string, error) {
...
+	hasRegion := region != ""
+	if notGlobal && hasRegion {
```
JoshuaWilkes, line 297:
```suggestion
	if !global && hasRegion {
```

JoshuaWilkes (changes requested):
I think there was a bug in the logic for global region

pkg/browsers/console.go
meyerjrr: Good catch, ill fix it up

pkg/browsers/console.go
```diff
@@ -284,9 +290,13 @@ func makeDestinationURL(service string, region string) (string, error) {
 
 	dest := prefix + serv + "/home"
 
-	//NOTE here: excluding iam here and possibly others as the region isnt in the uri of the webpage on the console
-	if region != "" || serv != "iam" {
+	//excluding region here if the service is apart of the global service list
+	//incomplete list of global services
+	_, Global := globalServiceMap[service]
```
JoshuaWilkes, line 295:
```suggestion
	_, global := globalServiceMap[service]
```

pkg/browsers/console.go
```diff
@@ -284,9 +290,13 @@ func makeDestinationURL(service string, region string) (string, error) {
...
+	hasRegion := region != ""
+	if !Global && hasRegion {
```
JoshuaWilkes, line 297:
```suggestion
	if !global && hasRegion {
```

JoshuaWilkes (changes requested):
Last nitpick here is the casing on this variable

JoshuaWilkes (approved):

PR #98, opened 2022-03-09
 build cleanup commands for tokens and config

### LINKED ISSUE #85: Add logout command
Made this it's own issue from #81. We should have a logout command (thinking `granted logout`) which clears SSO tokens.

meyerjrr:
- Added tokens command to: list tokens, remove tokens
- Added command to reset granted config (remove .granted folder items and clear environment variables)

Fixes #85


pkg/cfaws/secure_token_storage.go
```diff
@@ -28,8 +28,8 @@ func GetValidCachedToken(profileKey string) *SSOToken {
 }
 
 // Attempts to store the token, any errors will be logged to debug logging
-func StoreSSOToken(profileKey string, ssoTokenValue SSOToken) {
-	err := credstore.Store(profileKey, ssoTokenValue)
+func StoreSSOToken(profileKey string, ssoTokenValue SSOToken, profile string) {
```
JoshuaWilkes, line 31: because there may be more than one "root profile" with the same sso start url, this will overwrite the description every time you need to login if you login with a different root profile

pkg/credstore/credstore.go
```diff
@@ -50,6 +51,43 @@ func Clear(key string) error {
 	return ring.Remove(key)
 }
 
+func ClearWithProfileName(key string) error {
```
JoshuaWilkes, line 54: this will not work for all root profiles which share the same start URL, I would propose that you do this based on the start  URL rather than any profile name

pkg/granted/entrypoint.go
```diff
@@ -29,7 +29,7 @@ func GetCliApp() *cli.App {
 		UsageText:            "granted [global options] command [command options] [arguments...]",
 		Version:              build.Version,
 		HideVersion:          false,
-		Commands:             []*cli.Command{&DefaultBrowserCommand, &settings.SettingsCommand, &CompletionCommand},
+		Commands:             []*cli.Command{&DefaultBrowserCommand, &settings.SettingsCommand, &CompletionCommand, &DefaultTokenCommand, &DefaultClearCommand},
```
JoshuaWilkes, line 32: nitpick but these could just be called ClearCommand and TokenCommand

pkg/granted/tokens.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+		fmt.Fprintf(os.Stderr, "Granted is using %s. To change this run `granted browser set`.\n", conf.DefaultBrowser)
+
+		return nil
+	},
+}
+
+var TokenListCommand = cli.Command{
+	Name:  "list",
+	Usage: "Remove all saved tokens from keyring and delete all granted configuration",
```
JoshuaWilkes, line 32: this comment needs to reflect the command, looks like it was copied from the clear command

JoshuaWilkes (changes requested):

pkg/granted/tokens.go
```diff
@@ -0,0 +1,97 @@
...
+	Name:        "token",
+	Usage:       "Manage aws access tokens",
+	Subcommands: []*cli.Command{&TokenListCommand, &ClearTokensCommand, &ClearAllTokensCommand},
+	Action: func(c *cli.Context) error {
+		//return the default browser that is set
+		conf, err := config.Load()
+		if err != nil {
+			return err
+		}
+		fmt.Fprintf(os.Stderr, "Granted is using %s. To change this run `granted browser set`.\n", conf.DefaultBrowser)
```
JoshuaWilkes, line 25: Looks lik ethe action here is copied from the browser command

pkg/cfaws/assumer_aws_sso.go
```diff
@@ -54,17 +54,17 @@ func (c *CFSharedConfig) SSOLogin(ctx context.Context) (aws.Credentials, error)
 		return aws.Credentials{}, err
 	}
 	cachedToken := GetValidCachedToken(ssoTokenKey)
-	newToken := false
+	// newToken := false
```
JoshuaWilkes, line 57: Looks like this was commented out by accident

JoshuaWilkes (changes requested):

JoshuaWilkes:
@meyerjrr 

There is a way that you can make 2 column text work well in the select prompt

pkg/cfaws/assumer_aws_sso.go
meyerjrr: Not by accident. to update the descriptions I needed to redo a cache of the token to update it. 
I tried to make a lite version (so that we are not calling storessokey every time) that just updates the description of the key but the Ring package doesn't allow for updating keys just overriding.

JoshuaWilkes (approved):

PR #107, opened 2022-03-15
Fix logic for custom browser path and add stdout option

### LINKED ISSUE #105: Support machines with NO browsers installed better
Granted will persistently ask to set a default browser upon setup

We should allow users if they choose to skip setting a default browser

This should then fall back to returning a console url when passing a `-c` flag

[Example of current bad UX](https://shitposting.pictures/WaAjAF6mOz5u)

JoshuaWilkes:
Fixes custom browser paths not being saved to config
Fixes #105
Adds stdout and Firefox stdout options for browser defaults
Remove Hash from Firefox container name


pkg/browsers/browsers.go
```diff
@@ -25,7 +26,7 @@ var BravePathWindows = []string{`\Program Files\BraveSoftware\Brave-Browser\Appl
 var EdgePathMac = []string{"/Applications/Microsoft Edge.app/Contents/MacOS/Microsoft Edge"}
 var EdgePathLinux = []string{`/usr/bin/edge`}
 var EdgePathWindows = []string{`\Program Files (x86)\Microsoft\Edge\Application\msedge.exe`}
-var FirefoxPathMac = []string{"/Applications/Firefox.app/Contents/MacOS/firefox"}
+var FirefoxPathMac = []string{"/Applications/Firefox.app/Contents/MacOS/firefoxs"}
```
meyerjrr, line 29: Do we need to change this back? Im guessing you added the s for testing when its not installed

PR #115, opened 2022-03-22
Add config for keyring

chrnorm:
Adds optional config for the keyring secrets store that Granted uses to store cached AWS SSO credentials.

Example config (`~/.granted/config`):
```
DefaultBrowser = "FIREFOX"
[Keyring]
Backend="keychain"
```

The above config forces the `keychain` backend for keyring.

For any local testing of this PR, you'll need to update `~/.dgranted/config` rather than the release config file.


meyerjrr (approved):

JoshuaWilkes:
Another suggestion here is to set `keyring.Debug = true` when the verbose flag is used in the cli.

This will enable internal logging messages to be shown from keyring

PR #125, opened 2022-04-04
Add support for --exec flag in linux and mac

JoshuaWilkes:


pkg/assume/assume.go
```diff
@@ -26,6 +29,10 @@ func AssumeCommand(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
+
+	if assumeFlags.String("exec") != "" && runtime.GOOS == "windows" {
+		return fmt.Errorf("--exec flag is not currently supported on windows. If you would like this feature, consider opening an issue on our github repo to let us know")
```
chrnorm, line 34:
```suggestion
		return fmt.Errorf("--exec flag is not currently supported on Windows. Let us know if you'd like support for this: https://github.com/common-fate/granted/issues/new")
```

chrnorm:
@JoshuaWilkes I tested this as follows:
```
❯ dassume cf-dev --exec "aws s3 ls"

[cf-dev](ap-southeast-2) session credentials will expire 2022-04-05 14:20:05 +0100 BST


```

and didn't receive any output - am I using the `--exec` flag wrong?

I am a fan of [aws-vault's approach to command execution](https://github.com/99designs/aws-vault/blob/master/cli/exec.go#L358), where the command and arguments to execute are passed after a `--` string, avoiding having to write quotation marks around the subcommand. For us this might look like:

```
assume cf-dev exec -- aws s3 ls
```

With this approach you would not be able to use the role selector with `exec`, however I'd argue that this is not a problem as the `exec` flag is more likely to be used in shell scripts and commands where a user already knows the name of the profile.

JoshuaWilkes:
@chrnorm You used the command correctly, were you using Fish for this test?

Issue #122, opened 2022-03-28
Using Granted with `sops` gives a `SharedConfigErr`


chrnorm:
When assuming a role with Granted and then running [sops](https://github.com/mozilla/sops) the following error is returned:

```
Failed to get the data key required to decrypt the SOPS file.

Group 0: FAILED
  arn:aws:kms:<REGION>:<ACCOUNT>:key/mrk-<KMS_KEY_ID>: FAILED
    - | Error creating AWS session: SharedConfigErr: only one
      | credential type may be specified per profile: source
      | profile, credential source, credential process, web identity
      | token, or sso
```

Interestingly, sops works fine when using awsume and aws-sso-utils to access the role. SOPS is written in Go so perhaps we are introducing env variables which are affecting the AWS Go SDK.

PR #127, opened 2022-04-05
Refactor and simplify cred process assumer

### LINKED ISSUE #126: credential_process assumer does not use the AWS SDK method to fetch credentials
Based on feedback from Granted users in the slack channel.

I realised that Granted was not using the AWS SDK implementation of credential_process to get the credentials.

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/credentials@v1.11.2/processcreds

This leads to some users experiencing issues with using Granted with 3rd party credential processes

JoshuaWilkes:
Fixes #126 

Refactor and simplify the credentials process assumer implementation

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -61,7 +61,7 @@ jobs:
         uses: golangci/golangci-lint-action@v2
         timeout-minutes: 5
         with:
-          version: v1.39
+          version: latest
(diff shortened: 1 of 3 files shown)


chrnorm:
I think this will also fix #122 🙌

chrnorm (approved):

PR #130, opened 2022-04-12
Fix console session expires in 15 minutes + add session duration flag

### LINKED ISSUE #58: Increase lifetime of the session credential
Hey guys,

First of all, thank you for creating this tool! ❤️

My question is very simple, would it be possible to increase the lifetime of the session credential? By parameter it would be awesome!

For those who use the terminal during all day, it is a bit annoying to have to renew it all the time.

`session credentials will expire 2022-02-25 08:21:14 +0000 WET`

Cheers!

### LINKED ISSUE #128: Console session in firefox expires within 15min
For some reason when I open a console session for one of my chained profiles e.g.

`assume --console -s cloudformation example-prd`

After 15min the console session expires and I have to open another console session. For some reason, this happens with chained role sessions. This is an example of a config profile where the issue happens:

```
[profile example-sso]
sso_start_url=https://example.awsapps.com/start
sso_region=eu-west-1
sso_account_id= 012345678901
sso_role_name=example-support
region=eu-west-1

[profile example-prd]
role_arn=arn:aws:iam::012345678902:role/example-administrator
source_profile=example-sso
region=eu-central-1
```

Let me know if you need extra logs/details.

meyerjrr:
fixes #128 
fixes #58

### DIFF
--- a/pkg/cfaws/assumer_aws_sso.go
+++ b/pkg/cfaws/assumer_aws_sso.go
@@ -14,7 +14,6 @@ import (
 	ssooidctypes "github.com/aws/aws-sdk-go-v2/service/ssooidc/types"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/bigkevmcd/go-configparser"
-	"github.com/common-fate/granted/pkg/browsers"
 	"github.com/common-fate/granted/pkg/debug"
 	"github.com/fatih/color"
 	"github.com/pkg/browser"
@@ -26,11 +25,11 @@ type AwsSsoAssumer struct {
 }
 
 func (asa *AwsSsoAssumer) AssumeTerminal(ctx context.Context, c *CFSharedConfig, configOpts ConfigOpts) (aws.Credentials, error) {
-	return c.SSOLogin(ctx, browsers.BrowserOpts{}, configOpts)
+	return c.SSOLogin(ctx, configOpts)
 }
 
-func (asa *AwsSsoAssumer) AssumeConsole(ctx context.Context, c *CFSharedConfig, browserOpts browsers.BrowserOpts, configOpts ConfigOpts) (aws.Credentials, error) {
-	return c.SSOLogin(ctx, browserOpts, configOpts)
+func (asa *AwsSsoAssumer) AssumeConsole(ctx context.Context, c *CFSharedConfig, configOpts ConfigOpts) (aws.Credentials, error) {
+	return c.SSOLogin(ctx, configOpts)
 }
 
 func (asa *AwsSsoAssumer) Type() string {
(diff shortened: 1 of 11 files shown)


pkg/cfaws/assumer_aws_sso.go
```diff
@@ -103,6 +109,8 @@ func (c *CFSharedConfig) SSOLogin(ctx context.Context) (aws.Credentials, error)
 				if c.AWSConfig.MFASerial != "" {
 					aro.SerialNumber = &c.AWSConfig.MFASerial
 					aro.TokenProvider = MfaTokenProvider
+					aro.Duration = duration
```
JoshuaWilkes, line 112: this only applies the duration if an mfa serial is present on the role

pkg/cfaws/assumer_aws_sso.go
```diff
@@ -87,6 +87,12 @@ func (c *CFSharedConfig) SSOLogin(ctx context.Context) (aws.Credentials, error)
 	credProvider := &CredProv{rootCreds}
 
 	if requiresAssuming {
+		duration := time.Hour
+
+		if c.Opts.Duration > duration {
```
JoshuaWilkes, line 92: I feel like this should just apply the duration if it is specified

pkg/cfaws/profiles.go
```diff
@@ -8,10 +8,12 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/bigkevmcd/go-configparser"
+	"github.com/common-fate/granted/pkg/browsers"
 	"github.com/fatih/color"
 )
 
 type CFSharedConfig struct {
+	Opts browsers.BrowserOpts
```
JoshuaWilkes, line 16: My git feel is that this doesn't belong on the sharedconfig struct, as it is not specific to each profile.
My preference is to pass the duration into the assumers, make a dedicated Opts struct for the assumers.
And another for the Browser. this keeps our packages from having too much coupling

JoshuaWilkes (changes requested):

pkg/cfaws/profiles.go
```diff
@@ -152,10 +160,10 @@ func (c CFSharedConfigs) ProfileNames() []string {
 	return names
 }
 
-func (c *CFSharedConfig) AssumeConsole(ctx context.Context, args []string) (aws.Credentials, error) {
-	return AssumerFromType(c.ProfileType).AssumeConsole(ctx, c, args)
+func (c *CFSharedConfig) AssumeConsole(ctx context.Context, browserOpts browsers.BrowserOpts, configOpts ConfigOpts) (aws.Credentials, error) {
```
JoshuaWilkes, line 163: I think we should remove the browser opts param here and on all the assumer interfaces.

The assumers are intended to generate credentials.
The Browser package uses those credentials to get a browser session .


So they have seperate concerns and there should be no options in browserOpts which impact the the assumers

JoshuaWilkes (approved):

Issue #74, opened 2022-03-01
Panic on starting


d47zm3:
Error
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x4430458]

goroutine 1 [running]:
github.com/common-fate/granted/pkg/cfaws.(*uninitCFSharedConfig).init(0x46211c0, 0xc000213078)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/cfaws/profiles.go:88 +0x18
github.com/common-fate/granted/pkg/cfaws.(*uninitCFSharedConfig).init(0xc000110690, 0xc000213078)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/cfaws/profiles.go:98 +0x8e
github.com/common-fate/granted/pkg/cfaws.GetProfilesFromDefaultSharedConfig({0x47c6130, 0xc00019a010})
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/cfaws/profiles.go:71 +0x169
github.com/common-fate/granted/pkg/assume.AssumeCommand(0xc0001a7a80)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/assume/assume.go:28 +0x151
github.com/common-fate/granted/pkg/assume.GetCliApp.func2(0x2)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/assume/entrypoint.go:39 +0x19
github.com/common-fate/granted/pkg/updates.WithUpdateCheck.func1(0xc0001a7a80)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/updates/updates.go:69 +0x110
github.com/urfave/cli/v2.(*App).RunContext(0xc0002d44e0, {0x47c6130, 0xc00019a010}, {0xc00019c000, 0x2, 0x2})
        /Users/runner/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:322 +0x7a8
github.com/urfave/cli/v2.(*App).Run(...)
        /Users/runner/go/pkg/mod/github.com/urfave/cli/v2@v2.3.0/app.go:224
main.main()
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/cmd/assume/main.go:14 +0x45
```
Version (OS as well)
```
Version: 0.1.5
[...]

PR #131, opened 2022-04-12
Support profiles with (some) special characters

badideasforsale:
It appears that whatever bug this code referred to no longer exists as a problem. I tested assuming the profile named `bsherman-test-dest..,@#$%^&*()+=|}{:<>/?` without any errors. See attached screenshot for proof of this working.

success image
success image2


badideasforsale:
I wonder what changed between #74 and now to cause `.` or `@` characters to be allowable? I checked open issues but not closed ones before this PR.

console success

However, the Firefox plug-in doesn't seem thrilled about my choice of profile name:
console profile name not exactly happy but not complaining

.

JoshuaWilkes:
@badideasforsale Thanks for this PR, Something must have changed in the AWS SDK for Go as the errors were coming from that rather than our code, hence we added this warning.

PR #135, opened 2022-04-14
Adjust session duration implementation

JoshuaWilkes:
I found the the implementation of session duration had some issues.

One of which is that some AWS sso roles have a maximum session duration configured in AWS and cannot be extended. However when launching a browser session, we can specify a longer duration than the duration for the session credentials.

So in this PR I have added

`--duration` which configures the duration for session credentials in the terminal
`--browser-duration` which configures the browser session duration

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -112,7 +112,6 @@ func AssumeCommand(c *cli.Context) error {
 
 	configOpts := cfaws.ConfigOpts{Duration: time.Hour}
 	duration := assumeFlags.String("duration")
-	fmt.Fprintln(os.Stderr, duration)
 	if duration != "" {
 		d, err := time.ParseDuration(duration)
 		if err != nil {
@@ -120,8 +119,6 @@ func AssumeCommand(c *cli.Context) error {
 		}
 		configOpts.Duration = d
 	}
-
-	fmt.Fprintf(os.Stderr, "duration: %v\n", duration)
 
 	if len(assumeFlags.StringSlice("pass-through")) > 0 {
 		configOpts.Args = assumeFlags.StringSlice("pass-through")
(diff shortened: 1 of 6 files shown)


chrnorm:
Good pickup on browser session duration being a different value. 

AWS specifies a `duration_seconds` [config value](https://docs.aws.amazon.com/cli/latest/topic/config-vars.html) which can be added to roles in `~/.aws/config` - we should load this if it's provided on the role rather than forcing users to use a `--duration` argument. I have taken a look through the codebase and it looks like we don't support this.

Separately, I wonder whether we could allow users to add a field to `~/.aws/config` to configure the browser role duration, e.g. 
```
[profile test]
role_arn=arn:aws:iam:...
granted_browser_duration_seconds=3600
```

What do you think of this? My thinking here is that the values are specific for a particular role and won't change that often, so it is more cumbersome for a user to enter a `--browser-duration=1h --duration=1h` all the time.

Also, for instances where a user does pass an argument rather than us reading `~/.aws/config`, could we just use `--duration` in both cases, rather than `--duration` and `--browser-duration`? Is it possible for a user to use `assume --browser-duration=2h --duration=1h` and for this to actually work? (won't the IAM session itself expire once 1h has elapsed, even if the console has a longer session?)

meyerjrr:
Good suggestions @chrnorm! The focus for the duration flags fix was highlighted by this issue: #58 

I think the multiple flags might get confusing. Should there just be one flag that duration overrides both console session and the browser session?

JoshuaWilkes:
> 

@meyerjrr  I am running a test on this to verify that the browser session can be longer than the iam credentials.

JoshuaWilkes:
I realised a made some mistakes and have reverted to a single duration flag.

The console session cannot be longer than the credential session duration.
The default is that the browser session will be as long as the credentials used.

meyerjrr (approved):

PR #143, opened 2022-04-19
added settings option for ordering of profiles

### LINKED ISSUE #140: [Feature Request] Add ability to disable the Frecency ordering of profiles
As in the title, although I can see how Frecency ordering of profiles can be a desirable feature for a lot of users, it would be useful to have an option to disable it in the Granted configuration.

We name profiles via a convention (with a digit prefix) to ensure consistent ordering of the list of profiles. This allows us to minimize the possibility of accidentally assuming a role in the wrong account (due to muscle memory).

I realise that this is a low priority use case but would love to get your opinion on this feature.

Thanks for building a great productivity tool!

meyerjrr:
`granted settings set-order`
To choose either Frecency or Alphabetical for profile orderings when running `assume`

fixes #140

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -75,7 +75,17 @@ func AssumeCommand(c *cli.Context) error {
 
 	if profile == nil {
 
+		//load config to check frecency enabled
+		cfg, err := config.Load()
+		if err != nil {
+			return err
+		}
+
 		fr, profiles := awsProfiles.GetFrecentProfiles()
+
+		if cfg.Ordering == "Alphabetical" {
+			profiles = awsProfiles.GetAlphabeticalProfiles()
+		}
 		fmt.Fprintln(color.Error, "")
 		// Replicate the logic from original assume fn.
 		in := survey.Select{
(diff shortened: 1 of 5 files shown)


pkg/assume/assume.go
```diff
@@ -75,7 +75,17 @@ func AssumeCommand(c *cli.Context) error {
...
+		//load config to check frecency enabled
+		cfg, err := config.Load()
+		if err != nil {
+			return err
+		}
+
 		fr, profiles := awsProfiles.GetFrecentProfiles()
+
+		if cfg.Ordering == "Alphabetical" {
+			profiles = awsProfiles.GetAlphabeticalProfiles()
```
JoshuaWilkes, line 87: I'm thinking that this should just call awsProfiles.ProfileNames()

Rather than creating a new wrapper method here.

JoshuaWilkes (changes requested):

pkg/granted/settings/frecency.go
```diff
@@ -0,0 +1,64 @@
...
+
+var SetProfileOrderingCommand = cli.Command{
+	Name:  "set",
+	Usage: "Sets the method of ordering IAM profiles in the assume method",
+	Action: func(c *cli.Context) error {
+		cfg, err := config.Load()
+		if err != nil {
+			return err
+		}
+		withStdio := survey.WithStdio(os.Stdin, os.Stderr, os.Stderr)
```
JoshuaWilkes, line 40: Small nitpick , in the granted command
we don't need to use withStdio and we don't need to override stdout to stderr

This is because the `granted` command does not run with a wrapper script

JoshuaWilkes (approved):

PR #145, opened 2022-04-19
Truncate usernames getfederationtoken

### LINKED ISSUE #141: GetFederationToken 'Member must have length less than or equal to 32'
└❯ assume -c AssumeRole-$ACCOUNT_ID-$ROLE_NAME

> operation error STS: GetFederationToken, https response error StatusCode: 400, RequestID: ..., api error ValidationError: 1 validation error detected: Value 'Granted@AssumeRole-$ACCOUNT_ID-$ROLE_NAME' at 'name' failed to satisfy constraint: Member must have length less than or equal to 32

I see `aws-vault` had solved this by truncating. https://github.com/99designs/aws-vault/pull/377 

...but in my case since we've standardized around a large amount of common data in the front ie: `AssumeRole-$ACCOUNT_ID-...` I wonder if simply truncating would lead to collisions/issues; like if I wanted two or more granted console sessions open how would it handle if the only difference is at the `-$ROLE_NAME` section of that member name?...I'm guessing it wouldn't if it was truncated outright. And had me thinking instead of simply truncating, if possible, it would be better to compare against the existing list of credential "containers" as well as to come up with something unique and identifiable but also deal with truncating to a length of 32 or less.

meyerjrr:
fixes #141


pkg/cfaws/assumer_aws_iam.go
```diff
@@ -105,7 +105,13 @@ func getFederationToken(ctx context.Context, c *CFSharedConfig) (aws.Credentials
 	}
 
 	client := sts.NewFromConfig(cfg)
-	out, err := client.GetFederationToken(ctx, &sts.GetFederationTokenInput{Name: aws.String("Granted@" + c.Name), Policy: aws.String(allowAllPolicy)})
+	name := "Granted@" + c.Name
+
+	if len(name) > 32 {
+		name = name[0:32]
+	}
+	out, err := client.GetFederationToken(ctx, &sts.GetFederationTokenInput{Name: aws.String(name)})
```
JoshuaWilkes, line 113:
```suggestion
	out, err := client.GetFederationToken(ctx, &sts.GetFederationTokenInput{Name: aws.String(name), Policy: aws.String(allowAllPolicy)})
```

JoshuaWilkes (approved):

Issue #138, opened 2022-04-14
Add Support for other AWS Partitions such as GovCloud


fdamstra:
Right now, this works fine for me on my AWS commercial profiles, but fails for my govcloud profiles:

```
$ assume -c mdr-test-c2-gov

ℹ️  use -s to open a specific service (https://docs.commonfate.io/granted/usage/console)

Opening a console for mdr-test-c2-gov in your browser...
```

And then the browser gets the classic AWS "The credentials in your login link were invalid. Please contact your administrator.". This is because it's going to `signin.aws.amazon.com` instead of `signin.amazonaws-us-gov.com`.

I would imagine similar issues exist for the China partition and the top secret one.

I would imagine you could key off the region to determine which partition you were in.

jordiup:
Hi @fdamstra, thanks for checking in. It's awesome to hear Granted is being used in other partitions (like `aws-us-gov`). I've just opened a draft PR for this issue, we'll let you know when it's merged.

jordiup:
@fdamstra we would love it if you could test out the branch https://github.com/common-fate/granted/pull/146 locally

Our team doesn't have access to another AWS Partition so this would be a massive help

You will have to:
- Clone the repo
- Follow the [setup requirements](https://github.com/common-fate/granted/blob/main/CONTRIBUTING.md) (install Go)
- Run `make cli`
- Then instead of running assume, you will run **dassume**
- i.e. `dassume -c mdr-test-c2-gov`

fdamstra:
It does not work. Still opens the default signin.

I added a debug line in `pkg/browsers/console.go`:
```
  partition := GetPartitionFromRegion(region)
  fmt.Fprintf(color.Error, "\nPartition is detected as %s for region %s...\n", partition, region)
```

And it prints
```
Partition is detected as aws for region us-gov-east-1...
```

Maybe I can figure out enough go to find the problem. I'll look today.

fdamstra:
@jordiup , I reworked your PR into #150 . 

The issue was that the console URL needs to be changed based on the partition.

However, I also used a more dynamic method to determine the partition. You can decide whether you like it.

jordiup:
Nice work with the changes @fdamstra, your branch has now been reviewed and merged. I'll close this off 😄

PR #146, opened 2022-04-19
Add support for other aws partitions

jordiup:
Draft PR for #138


fdamstra:
This does not work for govcloud as the console URL is incorrect. I'm also concerned about you having to maintain a list of regions.

I've reworked this PR a little into #150 .

PR #150, opened 2022-04-20
Add support for other aws partitions by tokenizing regions

fdamstra:
Based off @jordiup 's work on adding partition support:
* Tokenizes the region and determines partition based on that. I think this will be more future proof than a static list of regions.
* Fixes the console URL, which differs by partition.

Tested and appears functional. Gets me a valid console in both govcloud and commercial. I am unable to test china or the ISOs.

Supercedes #146 to solve #138

### DIFF
--- a/pkg/browsers/console.go
+++ b/pkg/browsers/console.go
@@ -223,63 +223,33 @@ func (p PartitionHost) HostString() string {
 	return "signin.aws.amazon.com"
 }
 
-// https://news.ycombinator.com/item?id=23297872
-var AWS_REGIONS = []string{
-	"ap-south-1",
-	"eu-north-1",
-	"eu-west-3",
-	"eu-west-2",
-	"eu-west-1",
-	"ap-northeast-2",
-	"ap-northeast-1",
-	"me-south-1",
-	"ca-central-1",
-	"sa-east-1",
-	"ap-east-1",
-	"ap-southeast-1",
-	"ap-southeast-2",
-	"eu-central-1",
-	"us-east-1",
-	"us-east-2",
-	"us-west-1",
-	"us-west-2",
-}
-
-var AWS_CN_REGIONS = []string{
-	"cn-north-1",
-	"cn-northwest-1",
-}
-
-var AWS_ISO_REGIONS = []string{
-	"us-iso-east-1",
-}
-
-var AWS_ISO_B_REGIONS = []string{
-	"us-isob-east-1",
-}
-
-var AWS_US_GOV_REGIONS = []string{
-	"us-gov-west-1",
-	"us-gov-east-1",
+func (p PartitionHost) ConsoleHostString() string {
+	switch p {
+	case Default:
+		return "https://console.aws.amazon.com/"
+	case Gov:
+		return "https://console.amazonaws-us-gov.com/"
+	case Cn:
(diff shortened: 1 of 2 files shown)


pkg/browsers/console.go
```diff
@@ -281,8 +357,7 @@ func makeDestinationURL(service string, region string) (string, error) {
 		// and this avoids the need to keep the ServiceMap alphabetically sorted when developing Granted.
 		sort.Strings(validServices)
 
-		return "", fmt.Errorf("\nservice %s not found, please enter a valid service shortcut.\nValid service shortcuts: [%s]\n", service, strings.Join(validServices, ", "))
-
+		return "", fmt.Errorf("\nservice %s not found, please enter a valid service shortcut.\nValid service shortcuts: [%s]", service, strings.Join(validServices, ", "))
```
chrnorm, line 360: 👍 on this. Given that we're return an `err` here we don't need to worry about a newline at the end of the statement (I think the original implementation was a typo)

pkg/browsers/console.go
```diff
@@ -184,15 +184,90 @@ const (
...
+	switch p {
+	case Default:
+		return "https://console.aws.amazon.com/"
+	case Gov:
+		return "https://console.amazonaws-us-gov.com/"
+	case Cn:
+		return "https://console.amazonaws.cn/"
+	}
+	// Note: we're not handling the ISO and ISOB cases, I don't think they are supported by a public AWS console
+	return "console.aws.amazon.com"
```
chrnorm, line 236: Might be a typo here.

```suggestion
	return "https://console.aws.amazon.com/"
```

pkg/browsers/console.go
```diff
@@ -184,15 +184,90 @@ const (
...
+}
+
+func GetPartitionFromRegion(region string) PartitionHost {
```
chrnorm, line 239: Looks good. I am a fan of this approach!

chrnorm:
🎉 congratulations on the first contribution @fdamstra! This looks good - thanks a lot for helping Granted support other partitions. I've taken a quick look over the changes - @jordiup could you please have a look too?

pkg/browsers/console.go
fdamstra: This was @jordiup 's work from his original PR.

pkg/browsers/console.go
fdamstra: Good catch. Will update the PR.

pkg/browsers/console.go
```diff
@@ -184,15 +184,90 @@ const (
...
+
 	sessJSON, err := json.Marshal(sess)
+
 	if err != nil {
 		return "", err
 	}
 
+	partition := GetPartitionFromRegion(region)
+	// couldn't figure out how to flag this based on the '--verbose' setting
+	//fmt.Fprintf(color.Error, "Partition is detected as %s for region %s...\n", partition, region)
```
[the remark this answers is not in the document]
fdamstra: I've updated the code, adding the required first field of debug.Verbosity, but it doesn't output with the `--verbose` flag. Still not clear how this is supposed to work.

jordiup:
Thanks for the help @fdamstra, I'm really happy with the changes, you smashed it. I'll merge now ✅

PR #157, opened 2022-05-02
Adds --env flag to export credentials

JoshuaWilkes:
This proposal adds a --env flag, the purpose is to automatically export credentials to a .env file when assuming in the terminal.

This can be useful when using editors like vs-code with debugging features.

image

exports these to a .env
If the .env does not exist it will prompt to create one
```
	AWS_ACCESS_KEY_ID=
	AWS_SECRET_ACCESS_KEY=
	AWS_SESSION_TOKEN=
	AWS_REGION=
```


chrnorm (approved):
Looks good to me, and worked well when I tested the build locally.

Issue #161, opened 2022-05-12
[Feature Request] Use `duration_seconds` from `~/.aws/config` as default session duration


dsech:
Right now, as far as I can tell, the default session lasts 1 hour in console or browser. In order to change this I need to type
```
assume -c --duration 12h
```

aws-cli already has the setting `duration_seconds` when using IAM roles (https://docs.aws.amazon.com/cli/latest/topic/config-vars.html#using-aws-iam-roles) that controls how long the temporary tokens are valid.

It would be nice for **granted** to automatically use this setting as the default duration, so that I wouldn't have to type it in console every time.

I see this was suggested before in a PR comment here https://github.com/common-fate/granted/pull/135#issuecomment-1099159852

PR #166, opened 2022-05-25
added get duration from default profile

meyerjrr:
Assume will grab seconds_duration from default profile and assume role for that duration by default

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -121,6 +121,15 @@ func AssumeCommand(c *cli.Context) error {
 	}
 
 	configOpts := cfaws.ConfigOpts{Duration: time.Hour}
+
+	//attempt to get session duration from profile
+	defaultDuration, err := profile.DefaultDuration(c.Context)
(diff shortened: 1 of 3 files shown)


chrnorm:
Fixes #161

pkg/assume/assume.go
```diff
@@ -121,6 +121,15 @@ func AssumeCommand(c *cli.Context) error {
 	}
 
 	configOpts := cfaws.ConfigOpts{Duration: time.Hour}
+
+	//attempt to get session duration from profile
+	defaultDuration, err := profile.DefaultDuration(c.Context)
```
JoshuaWilkes, line 126: I propose that yuou can remove this fuctiuon call and jist use 
```suggestion
	if profile.AWSConfig.RuleDurationSeconds != nil {
	configOpts.Duration = *profile.AWSConfig.RuleDurationSeconds
	}
```

JoshuaWilkes (approved):

PR #168, opened 2022-05-26
Export creds

### LINKED ISSUE #164: [Feature request] Add flag to write credentials to ~/.aws/credentials under profile name headers.
My team is using Granted and love it, but also need to augment it with the [SSOcred utility](https://github.com/PredictMobile/aws-sso-credentials-getter) to copy temporary credentials to the user home AWS credentials path.

The `--env` flag is so close to what we need, but some of our tools use legacy connection methods and expect the credentials to be at `~/.aws/credentials ` and we can't update them to read from an `.env` instead.

It would be great to have a flag to export to that path and group credentials under the `[<profile_name>]` header.

meyerjrr:
Added flag for exporting creds to `~/.aws/credentials`

Fixes #164


pkg/cfaws/cred-exporter.go
```diff
@@ -0,0 +1,83 @@
...
+		fmt.Fprintln(color.Error, "Created file.")
+
+	}
+
+	credFile, err := configparser.NewConfigParserFromFile(credPath)
+	if err != nil {
+		return err
+	}
+
+	if credFile.HasSection(profileName) {
```
JoshuaWilkes, line 38: Can we replace this if with a simple

Remove section
Add section
Add items?

```suggestion
	err := credFile.RemoveSection("sectionname")
	if err != nil {
		return err
	}
	err = credFile.AddSection("sectionname")
	if err != nil {
		return err
	}
	err = credFile.Set(profileName, "aws_access_key_id", creds.AccessKeyID)
	if err != nil {
		return err
	}
	err = credFile.Set(profileName, "aws_secret_access_key", creds.SecretAccessKey)
	if err != nil {
		return err
	}
	err = credFile.Set(profileName, "aws_session_token", creds.SessionToken)
	if err != nil {
		return err
	}
	err = credFile.SaveWithDelimiter(credPath, "=")
	if err != nil {
		return err
	}
```

meyerjrr: From what I could tell in the docs, this is the only way of adding to the profile dict unfortunately.

https://pkg.go.dev/github.com/bigkevmcd/go-configparser#ConfigParser.Set

JoshuaWilkes: You may have misread my snippet

JoshuaWilkes (approved):

PR #169, opened 2022-05-27
Sso browser

meyerjrr:


pkg/cfaws/assumer_aws_sso.go
```diff
@@ -154,12 +154,26 @@ func SSODeviceCodeFlow(ctx context.Context, cfg aws.Config, rootProfile *CFShare
...
+	//check if sso browser path is set
+
+	config, err := grantCfg.Load()
 	if err != nil {
-		// fail silently
-		debug.Fprintf(debug.VerbosityDebug, color.Error, err.Error())
+		return nil, err
+	}
+	if config.CustomSSOBrowserPath != "" {
+		err = browsers.LaunchBrowserSession(url)
```
JoshuaWilkes, line 165: From what I can tell, launch browser session also includes logic to prompt to launch the url manually, though we have been adding these messages where we use the browser. 

In this section, we already print a message above to prompt the user to open the browser manually.

It also looks like you have left the default browser behaviour commented out. 

I think it should be in an else statement.


I propose that the LaunchBrowserSession function is renamed to OpenUrlWithCustomBrowser and it should behave the same way as browser.OpenUrl()

JoshuaWilkes (changes requested):

pkg/browsers/console.go
```diff
@@ -312,6 +312,33 @@ func MakeUrl(sess Session, opts BrowserOpts, service string, region string) (str
...
+func OpenUrlWithCustomBrowser(url string) error {
+
+	cfg, _ := config.Load()
+	if cfg == nil {
+		return browser.OpenURL(url)
+	}
+	if cfg.CustomSSOBrowserPath != "" {
+		cmd := exec.Command(cfg.CustomSSOBrowserPath, fmt.Sprintf(" %s ", url))
+		err := cmd.Start()
+		if err != nil {
```
JoshuaWilkes, line 324:
```suggestion
		return err
```

JoshuaWilkes: You can remove this message and leave it up to the calling function

JoshuaWilkes (approved):

PR #174, opened 2022-06-16
don't prefix with GrantedOutput if env var is set

chrnorm:
if NO_GRANTED_OUTPUT_PREFIX is true, don't put "GrantedOutput"
before the stdout of executed commands


nicgrayson:
Confirmed working for me

PR #177, opened 2022-06-22
Support arbitrary service names

alexjurkiewicz:
Allow users to use arbitrary names for AWS console services.

There's still an optional mapping step, useful for a) services that have unusual URLs, and b) nicknames.

### DIFF
--- a/pkg/browsers/console.go
+++ b/pkg/browsers/console.go
@@ -362,6 +362,7 @@ func makeDestinationURL(service string, region string) (string, error) {
 
 	serv := ServiceMap[service]
 	if serv == "" {
+		color.New(color.FgYellow).Fprintf(color.Error, "[warning] we don't recognize service %s but we'll try and open it anyway (you may receive a 404 page)", service)
 		serv = service
 	}
 
--- a/pkg/browsers/console.go
+++ b/pkg/browsers/console.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"os/exec"
 	"path"
-	"sort"
 	"strings"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -26,8 +25,14 @@ var ServiceMap = map[string]string{
 	"":               "console",
 	"ec2":            "ec2/v2",
 	"sso":            "singlesignon",
+ 	"ecs":            "ecs",
+	"eks":            "eks",
+	"athena":         "athena",
+	"cloudmap":       "cloudmap",
 	"c9":             "cloud9",
 	"cfn":            "cloudformation",
+	"cloudformation": "cloudformation",
+	"cloudwatch":     "cloudwatch",
 	"gd":             "guardduty",
 	"l":              "lambda",
 	"cw":             "cloudwatch",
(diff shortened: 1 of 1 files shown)


pkg/browsers/console.go
```diff
@@ -376,17 +362,7 @@ func makeDestinationURL(service string, region string) (string, error) {
...
-			validServices = append(validServices, s)
-		}
-		// present the strings in alphabetical order.
-		// Yes, this is a bit of computation - but our arrays are quite small
-		// and this avoids the need to keep the ServiceMap alphabetically sorted when developing Granted.
-		sort.Strings(validServices)
-
-		return "", fmt.Errorf("\nservice %s not found, please enter a valid service shortcut \nValid service shortcuts: [%s]\n", service, strings.Join(validServices, ", "))
-
+		serv = service
```
chrnorm, line 365: I think it's worth displaying a warning to the user if we don't have the service in our `ServiceMap` mapping. I know this creates more maintenance burden for us but I think it improves the UX for users if they make an accidental typo e.g. `assume -c -s ecc2`.

```suggestion
		color.New(color.FgYellow).Fprintf(color.Error, "[warning] we don't recognize service %s but we'll try and open it anyway (you may receive a 404 page)", service)
		serv = service
```

If you make a typo you get a rather unhelpful 404 page, which might make users think there is an issue with the AWS console rather than a mistyped service.

image

pkg/browsers/console.go
```diff
@@ -26,14 +26,8 @@ var ServiceMap = map[string]string{
 	"":               "console",
```
chrnorm, line 26: GitHub won't let me suggest a change as the imports aren't changed in this file, but you'll need to remove the `"sort"` import as it's no longer used and causes the build to break (Go is very strict on unused imports)

pkg/browsers/console.go
```diff
@@ -26,14 +26,8 @@ var ServiceMap = map[string]string{
...
 	"ec2":            "ec2/v2",
 	"sso":            "singlesignon",
-	"ecs":            "ecs",
```
chrnorm, line 29: In line with my suggestion below I'd propose reverting the deleted services from `ServiceMap` and leaving the map unchanged. We can keep adding known services to this so that the only times we're falling back to an arbitrary name is (a) when a new service is introduced that we don't yet handle and (b) when a user has made a typo in the service name.

chrnorm (changes requested):
Thanks so much for the PR @alexjurkiewicz - this makes Granted far more usable for services that we don't have handled in `ServiceMap`!

The branch currently fails to compile for me locally due to a pesky unused import - I've added a review comment with instructions on how to fix.

Additionally, I'm confident that long-term on this project we can maintain a list of AWS service names that correspond to correct console URLs. Because of this, I've proposed that we log a warning message if the service name doesn't match our `ServiceMap` mapping, with the idea that this could help users if they make a typo in the service name.

pkg/browsers/console.go
alexjurkiewicz: done!

alexjurkiewicz:
Thanks for the feedback, I've made your suggested changes 🙌

chrnorm (approved):
🙌 🙌  thanks @alexjurkiewicz!

PR #179, opened 2022-07-01
assume: Support short region names (eg 'use1')

alexjurkiewicz:
This also requires an update to [the docs](https://docs.commonfate.io/granted/usage/console/#opening-the-console-with-a-specific-region), but I couldn't find which repo they live in...


pkg/browsers/console_test.go
```diff
@@ -0,0 +1,51 @@
...
+		wantErr bool
+	}{
+		{"<blank>", args{""}, "us-east-1", false},
+		{"us-east-1", args{"us-east-1"}, "us-east-1", false},
+		{"ue1", args{"ue1"}, "us-east-1", false},
+		{"ue", args{"ue"}, "us-east-1", false},
+		{"afs1", args{"afs1"}, "af-south-1", false},
+		{"apse3", args{"apse3"}, "ap-southeast-3", false},
+		{"cc", args{"cc"}, "ca-central-1", false},
+		{"cc1", args{"cc"}, "ca-central-1", false},
```
JoshuaWilkes, line 22:
```suggestion
		{"cc1", args{"cc1"}, "ca-central-1", false},
```

pkg/browsers/console_test.go
```diff
@@ -0,0 +1,51 @@
...
+		{"ec1", args{"ec1"}, "eu-central-1", false},
+		{"euc1", args{"ec1"}, "eu-central-1", false},
```
JoshuaWilkes, line 24:
```suggestion
		{"euc1", args{"euc1"}, "eu-central-1", false},
```

JoshuaWilkes (changes requested):
Thanks for the PR!

I have flagged some typos in the test cases, I will get a full review done later in the week.

Overall I like the idea this shorthand syntax

JoshuaWilkes (commented):
Thanks for your PR.
What I'm thinking for this is that I would like to see the expandRegion function shifted out into the cfaws package.

The region transformation should be applied in the assume command when the assumeflags.String("region") is read.

I think it will be best to apply this transformation as close as possible to where we read the flag so that we can always treat "region" in the codebase as a fully formed region. Otherwise I think it will be easy to introduce regressions as its not clear that there is a difference between a shorthand region and a fully formed region.

We can also support setting the region in the terminal via the flag as well, as it currently is not the case

I will apply the following changes and get this merged in
- Shift expandRegion and the test into the cfaws package as region.go and region_test.go
- in pkg/assume/entrypoint.go update the region check to expand the region if the flag is supplied

```
	var region string
	// The region flag may be supplied in shorthand form, first check if the flag is set and expand the region
	// else use the profile region
	if assumeFlags.String("region") != "" {
		regionFlag := assumeFlags.String("region")
		region, err = cfaws.ExpandRegion(regionFlag)
                if err != nil {
			return err
		}
	} else {
		region, _, err = profile.Region(c.Context)
		if err != nil {
			return err
		}
	}
```

Issue #117, opened 2022-03-23
[Feature Request] Granted should support assuming profiles that are only defined in the .credentials file


JoshuaWilkes:
Some of our users have reported that they use profiles which are defined only in the credentials file, rather than in config.

Currently, to use granted with these profiles they need to be added into the config file.

TryTryAgain:
Similarly, I found that granted was very specific with how each was prepended or absent of `profile`

I ended up needing to have `[profile AssumeRole-SomeAccountID-SomeRole]` in the ~/.aws/config whereas in ~/.aws/credentials I needed to have it without "profile" `[AssumeRole-SomeAccountID-SomeRole]` otherwise it wasn't working. We also only had things stuffed into our ~/.aws/credentials file originally.

PR #182, opened 2022-07-09
Assume profiles from credentials file [https://github.com/common-fate/granted/issues/117]

TasSM:
This pull request addresses feature request https://github.com/common-fate/granted/issues/117

Granted will now look for any additional profiles defined only in the ~/.aws/credentials file which are not defined in the default config file. If a profile with the same name is defined in the default config file then the profile in the credentials file will be ignored.

Tested locally with 2 different profiles defined across both files and appears to work as expected.


JoshuaWilkes (approved):

JoshuaWilkes:
Looks good to me!

PR #189, opened 2022-07-18
Make granted sessions uniquely identified

### LINKED ISSUE #163: [Feature Request] The name of the federated ID should have unique identifier
When using `assume`, the name of the IAM entity is `arn:aws:sts::1234567890:assumed-role/MyRole/Granted-{{ profile_name }}`. The problem with this is it makes it really hard to audit. We have a template for all user's to follow for their AWS config set up and it's an auditing nightmare to see who did what/when.

AWS has a timestamp or some unique integer in the name (example: `arn:aws:sts::1234567890:assumed-role/MyRole/botocore-session-1295020300`).

JoshuaWilkes:
Fixes #163
Use a unique ID for the session name

### DIFF
--- a/pkg/cfaws/session_name.go
+++ b/pkg/cfaws/session_name.go
@@ -1,0 +1,9 @@
+package cfaws
+
+import "github.com/segmentio/ksuid"
+
+// sessionName returns a unique session identifier for the aws console
+// this ensures that user activity can be easily audited per session
+func sessionName() string {
+	return "gran-" + ksuid.New().String()
+}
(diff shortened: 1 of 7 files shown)


pkg/cfaws/session_name.go
```diff
@@ -0,0 +1,10 @@
+package cfaws
+
+import "github.com/segmentio/ksuid"
+
+// sessionName returns a unique session identifier for the aws console
+// this ensures that user activity can be easily audited per session
+// this uses the convenient ksuid library for generating unique IDs
+func sessionName() string {
+	return "gran-" + ksuid.New().String()
```
meyerjrr, line 9: Should this be "granted"?

JoshuaWilkes: I needed the name to come in under a character limit, maybe I should just use a different UUID format.
I think I will swap out the KSUID for something else

Issue #184, opened 2022-07-13
Combining aws sso and assumeRole not working


shotty1:
I am currently trying to replace awsu.me with granted.dev for me and my team but am hitting one major roadblock.
In my scenario I need to log in to an account with AWS SSO, and then assume a role in another account.

We got this running by combining awsu.me with ben Kehoes aws-sso-util (see config below)

But when trying to achieve the same thing with granted.dev I get the following error (tried with the -ex parameter too)

```bash
assume --verbose target-account
DEBUG: starting update check
2022/07/13 17:23:46 [keyring] Considering backends: [keychain pass file]
2022/07/13 17:23:46 [keyring] Querying keychain for service="granted", account="https://d-1234567890.awsapps.com/start", keychain="login.keychain"
2022/07/13 17:23:46 [keyring] Found item "granted"
operation error STS: AssumeRole, https response error StatusCode: 403, RequestID: somerequestid, api error AccessDenied: User: arn:aws:sts::111111111111:assumed-role/AWSSSOASSUMEDROLENAME is not authorized to perform: sts:AssumeRole on resource: arn:aws:iam::222222222222:role/TARGETROLENAME
```

Looging into the sso-login-account works without problems

Config parts:

```bash
[profile soo-login-account]
credential_process=aws-sso-util credential-process --profile soo-login-account
sso_start_url=https://d-1234567890.awsapps.com/start
sso_region=eu-central-1
sso_account_id=111111111111
sso_role_name=AWSSSOROLENAME
region=eu-central-1

[profile target-account]
role_arn=arn:aws:iam::222222222222:role/TARGETROLENAME
source_profile=soo-login-account
external_id=mysupersecretexternalid
```

granted version: 0.2.2

JoshuaWilkes:
Hey @shotty1 

Does the external_id play a part in your cross account trust policy?

Granted does not currently use external_id so this may well be your issue.
This is the guide I looked at https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html

I imagine if we added this into our requests that your use-case would work

JoshuaWilkes:
@shotty1 If you have time, you can try building granted from source on this branch and re-testing your usecase
https://github.com/common-fate/granted/pull/187

Otherwise we can test when we get a chance

shotty1:
Hey, yes the external Id plays a role. Will try to test later and get back to you.

shotty1:
works like a charm. Thanks a lot!
[...]

PR #192, opened 2022-07-18
reduce loading times by deferring initialisation of profile

### LINKED ISSUE #172: Slow with large number of AWS Profiles
Hi,

I manage a large number of AWS profiles due to managing a few thousand AWS accounts. My `~/.aws/config` file is 7000 lines long. Every call to `assume` takes 10-20 seconds on my macbook pro 16inch M1 Pro laptop. 

Not sure if there is something I can configure to reduce the lookup times.

JoshuaWilkes:
Fixes #172 

This PR improves the performance of the assume command by deferring initialisation of an AWSProfile parent tree until after a profile is selected.

In my testing with 150 profiles, this reduced the time to prompt in the assume command from 1.7s to 7ms as it is effectively just a few file reads.

The effect is that autocomplete and assume with no profile should take ms to load rather than seconds for large configs.
Only the selected profile will be parsed for type and parents before assuming.

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -15,7 +15,6 @@ import (
 	"github.com/common-fate/granted/pkg/browsers"
 	"github.com/common-fate/granted/pkg/cfaws"
 	"github.com/common-fate/granted/pkg/config"
-	"github.com/common-fate/granted/pkg/debug"
 	"github.com/common-fate/granted/pkg/testable"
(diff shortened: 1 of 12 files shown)


JoshuaWilkes:
also addresses an issue with #184 where region is not inherited from parents if it is not set on a child profile

meyerjrr (approved):

PR #197, opened 2022-07-19
adding check for credentials file not existing before erroring out

dcarbone:
Currently with 0.2.3+, `assume` will always error out if you do not have an `~/.aws/credentials` file present despite it not being required.

This patch tests the error for [os.ErrNotExist](https://github.com/golang/go/blob/master/src/os/error.go#L24) by way of the [os.IsNotExist](https://github.com/golang/go/blob/master/src/os/error.go#L92) err utility func.

### DIFF
--- a/pkg/cfaws/profiles.go
+++ b/pkg/cfaws/profiles.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 	"sort"
 	"strings"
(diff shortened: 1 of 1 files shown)


chrnorm:
@dcarbone thanks for your contribution! I added an additional check on top of your branch in #198 and have merged this 👍

dcarbone:
oop!  sorry i missed that one!

Issue #201, opened 2022-07-22
The -s flag should imply --console


chrnorm:
From the Common Fate Community Slack:

> I don’t think using -s by itself does anything, I don’t see why I can’t just do assume -s iam foobar-users-acct or even assume -c iam foobar-users-acct

I think our initial implementation here can be defaulting to opening the console if `assume -s iam <profile>` is used to avoid the need for the `-c` flag to be provided.

cmddx:
Done and done. Fixed in #204 which has now been merged into main.

PR #204, opened 2022-07-26
The -s flag now assumes -c

cmddx:
This is a fix for issue #201.
Boolean logic for opening browser now checks that service flag length is > 0
Updated help page to clarify what --c does

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -147,7 +147,7 @@ func AssumeCommand(c *cli.Context) error {
 		configOpts.Args = assumeFlags.StringSlice("pass-through")
 	}
 
-	openBrower := !assumeFlags.Bool("env") && (assumeFlags.Bool("console") || assumeFlags.Bool("active-role") || len(assumeFlags.String("service")) > 0 || assumeFlags.Bool("url"))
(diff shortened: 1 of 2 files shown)


pkg/assume/assume.go
```diff
@@ -147,7 +147,7 @@ func AssumeCommand(c *cli.Context) error {
 		configOpts.Args = assumeFlags.StringSlice("pass-through")
 	}
 
-	openBrower := !assumeFlags.Bool("env") && (assumeFlags.Bool("console") || assumeFlags.Bool("active-role") || assumeFlags.Bool("url"))
+	openBrower := !assumeFlags.Bool("env") && (assumeFlags.Bool("console") || assumeFlags.Bool("active-role") || len(assumeFlags.String("service")) > 0 || assumeFlags.Bool("url"))
```
JoshuaWilkes, line 150: Nitpick: I like to use a not empty check if the any length non empty string should be true as I find it reads easier.

Buts its exactly the same check in the end

```suggestion
	openBrower := !assumeFlags.Bool("env") && (assumeFlags.Bool("console") || assumeFlags.Bool("active-role") || assumeFlags.String("service") != "" || assumeFlags.Bool("url"))
```

JoshuaWilkes (approved):

Issue #205, opened 2022-07-28
AWS_SESSION_EXPIRATION gives different timezones for different profiles


sosheskaz:
When assuming two different profiles:

```bash
$ assume profile1; echo $AWS_SESSION_EXPIRATION

[profile1](us-east-1) session credentials will expire 2022-07-29 04:16:33 -0500 CDT
2022-07-29T04:16:33-05:00

$ assume profile2; echo $AWS_SESSION_EXPIRATION

[profile2](us-east-1) session credentials will expire 2022-07-28 17:16:51 -0500 CDT
2022-07-28T22:16:51Z
```

Here's my AWS config:

```ini
[profile profile1]
cli_pager =
sso_start_url = https://<sso>.awsapps.com/start
sso_region = us-west-2
sso_account_id = <acct>
sso_role_name = profile1

...

[profile profile2]
source_profile = profile1
role_arn = ...
```

My desired behavior would be for the timestamp to be consistent. Either always system time (UTC-5 in my case) or always UTC time.

PR #206, opened 2022-07-29
Localize datetime for session token expiration

sosheskaz:
This fixes https://github.com/common-fate/granted/issues/205 by forcing session tokens to always use local time when writing output from the assume command.

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -195,7 +195,7 @@ func AssumeCommand(c *cli.Context) error {
 		sessionExpiration := ""
 		green := color.New(color.FgGreen)
 		if creds.CanExpire {
-			sessionExpiration = creds.Expires.Format(time.RFC3339)
+			sessionExpiration = creds.Expires.Local().Format(time.RFC3339)
 			if os.Getenv("GRANTED_QUIET") != "true" {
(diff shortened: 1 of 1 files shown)


pkg/assume/assume.go
```diff
@@ -195,7 +195,7 @@ func AssumeCommand(c *cli.Context) error {
 		sessionExpiration := ""
 		green := color.New(color.FgGreen)
 		if creds.CanExpire {
-			sessionExpiration = creds.Expires.Format(time.RFC3339)
+			sessionExpiration = creds.Expires.Local().Format(time.RFC3339)
 			if os.Getenv("GRANTED_QUIET") != "true" {
 				green.Fprintf(color.Error, "\n[%s](%s) session credentials will expire %s\n", profile.Name, region, creds.Expires.Local().String())
```
sosheskaz, line 200: Should this use `sessionExpiration` instead of `creds.Expires.Local().String()` for consistency/simplicity?

meyerjrr: Thanks for opening the PR! LGTM. If you want to make this change for consistency, we can merge it in.

chrnorm (approved):

chrnorm:
thanks @sosheskaz!

PR #218, opened 2022-08-17
Fix sso aws outputs

### LINKED ISSUE #212: Specifying SSO variables via the --sso command should not set AWS_PROFILE
If `AWS_PROFILE` is set, AWS CLI commands all fail with an error similar to the below

```
The config profile (GrantedAdministratorAccess) could not be found
```

Done = when --sso flag is used, all env vars **except** AWS_PROFILE should be set. AWS_PROFILE should not be set at all (not even to an empty string, as this also causes the CLI to throw errors).

meyerjrr:
Fixes #212 
Changes scripts to not output AWS_PROFILE when assuming using -sso


scripts/assume.bat
```diff
@@ -48,5 +48,32 @@ for /f "tokens=1,2,3,4,5,6,7 delims= " %%a in ("%ASSUME_OUTPUT%") do (
 
         Exit /b %ASSUME_STATUS%
     )
+
+    if "%%a" == "GrantedAssumeSSO" (
```
JoshuaWilkes, line 52: @meyerjrr This looks like it would work, however the way the script is written, if you put the value "None" in place of the aws_profile output, it will not be set.

pkg/assume/assume.go
```diff
@@ -246,8 +246,14 @@ func AssumeCommand(c *cli.Context) error {
 		// DO NOT REMOVE, this interacts with the shell script that wraps the assume command, the shell script is what configures your shell environment vars
 		// to export more environment variables, add then in the assume and assume.fish scripts then append them to this output preparation function
 		// the shell script treats "None" as an emprty string and will not set a value for that positional output
-		output := PrepareStringsForShellScript([]string{creds.AccessKeyID, creds.SecretAccessKey, creds.SessionToken, profile.Name, region, sessionExpiration})
-		fmt.Printf("GrantedAssume %s %s %s %s %s %s", output...)
+		if c.Bool("sso") {
```
JoshuaWilkes, line 249: Instead of this,

You can use an empty string for profile.Name if it is sso then it will be set to "None" which in turn does not export the env var

JoshuaWilkes (changes requested):
This looks like it will work, however I have requested changes to use the existing implementation instead which is a very straightforward change to the final output

Issue #222, opened 2022-08-19
--exec doesn't pass environment variables to the child shell


chrnorm:
Raised in #216 - this is a separate issue to whitespace handling so opening a new issue to track work on this.

Currently environment variables are not passed to the command executed with `--exec`.

sosheskaz:
Thanks for opening and confirming the issue; created a PR with my suggested fix.

PR #223, opened 2022-08-19
Propagate environment vars to exec command

sosheskaz:
Fixes https://github.com/common-fate/granted/issues/222

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -280,10 +280,9 @@ func RunExecCommandWithCreds(cmd string, creds aws.Credentials, region string) e
 	fmt.Print(MakeGrantedOutput(""))
 	args := strings.Split(cmd, " ")
 	c := exec.Command(args[0], args[1:]...)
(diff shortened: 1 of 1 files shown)


sosheskaz:
Validated by comparing the output before and after this change:

```
FORCE_NO_ALIAS=true go run ./cmd/assume my-profile --exec printenv
```

Before, it would print only the AWS variables that are explicitly loaded into the environment; now it lists all env vars.

sosheskaz:
If the same env key is specified twice in the `Environ` slice, e.g. `[..., "AWS_REGION=us-east-1", ..., "AWS_REGION=us-west-2"]`, then it will use the latter value.

From the godocs: https://pkg.go.dev/os/exec#Command

> // If Env contains duplicate environment keys, only the last
> // value in the slice for each duplicate key is used.

So this should maintain the desired behavior of Granted's env vars taking precedence, by virtue of being specified later.

JoshuaWilkes (approved):
Thanks for the fix!

Issue #129, opened 2022-04-08
[Feature Request] Support automatic refresh of temporary session credentials 


dannysteenman:
A nice feature to have is the ability to rotate temp session credentials automatically instead of manually triggering the assume command again.

JoshuaWilkes:
Closing this for now as it is implemented for ZSH checkout the documentation here https://docs.commonfate.io/granted/recipes/automatically_reassume

PR #228, opened 2022-08-25
Add automatic reassume functionality for zsh

sosheskaz:
Partially fixes issue #129. Scope limited to zsh.

This script is crude, and has several shortcomings, but the solutions are nontrivial and it should work as-is.

I previously implemented this as a profile, so it has some rough edges from the conversion. Since the hook should only be added once, it needed extra logic to test that, which I have not yet scrutinized.

The date logic is tricky, and limited by the need to maintain compatibility with both GNU and BSD `date` implementations.

### DIFF
--- a/scripts/assume
+++ b/scripts/assume
@@ -12,7 +12,7 @@ if [[ "${BASH_SOURCE[0]}" != "${0}" ]]; then
 fi
 
 # in some cases such as github actions, the SHELL env var is not available to the program
-# so here we set it explicitly 
+# so here we set it explicitly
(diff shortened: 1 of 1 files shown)


sosheskaz:
I've been using as my `assume` script this for the past week on a macbook, and it has been reliable.

chrnorm:
🙌🙌 thanks so much for this contribution @sosheskaz! I've just tested and verified the functionality working on ZSH on my own machine. We will add a recipe for this into our [docs](https://docs.commonfate.io/granted).

chrnorm (approved):

PR #232, opened 2022-08-30
config file flags

### LINKED ISSUE #229: env variable AWS_CONFIG_FILE is ignored
I work with a lot of AWS accounts and often I use `AWS_CONFIG_FILE` environment variable to logically separate my config files so I don't have one giant messy one. `aws-vault`, for example, picks this up and reads from there. It helps when giving demo's/presentations not to show the other profiles that I have.

https://github.com/99designs/aws-vault/blob/9a931236194799be720676abf17889ad7d3ec53b/vault/config.go#L39

JoshuaWilkes:
Fixes #229

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -46,7 +46,10 @@ func AssumeCommand(c *cli.Context) error {
 		activeRoleFlag := assumeFlags.Bool("active-role")
 
 		withStdio := survey.WithStdio(os.Stdin, os.Stderr, os.Stderr)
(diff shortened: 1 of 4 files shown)


chrnorm (changes requested):

stedrow:
@JoshuaWilkes is this being implemented somewhere else? I'd love this functionality from granted.

JoshuaWilkes:
@stedrow We will be implementing this!
Recently we have been working on improvements to config parsing and switched to a different package in #297 
We will be able to revisit supporting config file flags soon now that this is done

PR #233, opened 2022-08-31
Allow filtering by multiple space-separated tokens

alexjurkiewicz:
I use profile names of the form `product_env_rolename_region`, so it's
often useful to search for something like "dev us-east-1". Since spaces
aren't allowed in profile names, this should be completely
backwards-compatible.


chrnorm:
🙌 just tested this and found it extremely user friendly. Thanks so much for the contribution @alexjurkiewicz!

chrnorm (approved):

Issue #84, opened 2022-03-03
Firefox quits if you exit the terminal that launched it


JoshuaWilkes:
granted v0.1.5

To Reproduce 
1. Set browser to firefox
2. assume -c profile
3. ctrl + c in the terminal which appears to be free
4. firefox will close

PR #252, opened 2022-09-13
refactor browser launching logic and use better process forking

chrnorm:
`pkg/browser` has become a bit messy. The module contains code for several different purposes, including:
- browser path detection
- browser configuration and setup wizards
- launching browsers
- handling the AWS federated auth login flow to get a browser URL

In addition to this there are a bunch of nested function calls and control flow which is difficult to understand, as well as enums that are only being used in one or two places.

This PR shifts browser launching logic to `pkg/launcher` and removes a lot of the nested function calls which were making the 'assume' command code difficult to read. Logic relating to retrieving console URLs via federated login is moved to `pkg/console`.

Additionally, it implements better process forking, so that pressing Control+C in a terminal after you've run `assume -c` will no longer close your browser window.

This will require acceptance testing across MacOS/Linux/Windows to verify it hasn't caused any regressions, as I've changed the structure of things when I've simplified the code. Manually tested and working well on MacOS with Firefox, Stdout, and FirefoxStdout browsers.


chrnorm:
Fixes #84

pkg/launcher/chrome_profile.go
```diff
@@ -0,0 +1,35 @@
...
+		l.ExecutablePath,
+		"--user-data-dir=" + l.UserDataPath,
+		"--profile-directory=" + profileName,
+		"--no-first-run",
+		"--no-default-browser-check",
+		url,
+	}
+}
+
+func chromeProfileName(profile string) string {
```
chrnorm, line 29: @meyerjrr I've kept this as-is from the previous implementation. Is there any reason we are using a hash, rather than just the name of the profile itself here?

meyerjrr: Was to help with the container functionality from memory, there was an issue with overlapping names. Will test it out

meyerjrr (approved):

pkg/launcher/chrome_profile.go
chrnorm: No worries, I've left it all the same as previous to avoid breaking changes. In future it could be good to see if we could just use the profile name here

PR #253, opened 2022-09-14
CLI-11 Add ability to use plain text aws sso token from aws/sso/cache dir if available. 

Eddie023:
### Dependent PR
1. https://github.com/common-fate/granted/pull/256

### Description
1. Add ability to use plain text aws sso token if available. 
2. `credential_process` will automatically redirect users to SSO login flow if valid token is not found for `granted` prefix profiles. (custom granted AWS config profiles)

### Summary of changes
1. When initializing profiles now, check if the AWS config file has `granted` prefix. This is necessary because aws `loadSharedConfig` func fails if it doesn't have default aws config. 
2. If `granted` prefix is available then convert values to default AWS config  values & create a custom credential provider that will retrieve necessary sso access_token & sso session_token from sso cache directory.
3. Add ability to dump plain text SSO token in default SSO cache directory.


pkg/cfaws/profiles.go
```diff
@@ -35,6 +35,8 @@ type Profile struct {
 	AWSConfig    config.SharedConfig
 	Initialised  bool
 	LoadingError error
+	// set to true if aws temp credentails are fetched from ~/.aws/sso/cache dir.
+	IsDefaultCredential bool
```
chrnorm, line 39: can we call this something like `UsePlaintextSSOToken`? Just thinking that "default" is a bit confusing, as Granted CLI defaults to using the keychain, whereas "default" in this context means "don't use the keychain".

```suggestion
	UsePlaintextSSOToken bool
```

pkg/granted/credentialsprocess.go
```diff
@@ -30,12 +31,14 @@ var CredentialsProcess = cli.Command{
...
-	Flags:       []cli.Flag{&cli.StringFlag{Name: "profile"}, &cli.StringFlag{Name: "rule"}},
+	Flags:       []cli.Flag{&cli.StringFlag{Name: "profile", Required: true}, &cli.StringFlag{Name: "rule"}},
 	Action: func(c *cli.Context) error {
 
 		url := ""
 		profileName := c.String("profile")
-		// ruleId := c.String("rule")
+		if !strings.HasPrefix(profileName, "granted.") {
+			log.Fatal("Profile name must start with 'granted.' prefix. Check your config file to verify.")
+		}
```
chrnorm, line 41: can we remove this check? if this function is being called, it means that the user has correctly set up their credential process, so it doesn't matter what their profile is named. This forces advanced users who might want to name their profiles a different way to use our `granted.` naming scheme.

```suggestion
```

pkg/granted/credentialsprocess.go
```diff
@@ -46,62 +49,51 @@ var CredentialsProcess = cli.Command{
...
-		if profileName == "" {
-			log.Fatalln("Mis-configured aws config file.\n--profile flag must be passed")
-		}
-
-		var profile *cfaws.Profile
-
 		profiles, err := cfaws.LoadProfiles()
 		if err != nil {
-			return err
+			log.Fatal(err)
```
chrnorm, line 54: Is there any specific need here to use log.Fatal? If we return `err`, `urfav/cli` should log the error for us!

```suggestion
			return err
```

chrnorm (changes requested):
Thanks for this @Eddie023! I've left some review comments.

A quick question, are we using this schema for AWS profiles?
```
[profile admin]
granted_sso_start_url = https://example.awsapps.com/start
granted_sso_account_name = Testing
granted_sso_account_id = 123456789012
granted_sso_role_name = GrantedAdministratorAccess
region = ap-southeast-2
credential_process = granted credential-process --profile admin --url https://granted.example.com
```

From my review of the PR, it appears that we are using
```
[profile admin]
sso_start_url = https://example.awsapps.com/start
sso_region = ap-southeast-2
sso_account_name = Testing
sso_account_id = 123456789012
sso_role_name = GrantedAdministratorAccess
region = ap-southeast-2

[profile granted.admin]
credential_process = granted credential-process --profile admin --url https://granted.example.com
```

This leads to an undesired effect as users might use `aws s3 ls --profile admin` and then they receive the same error we're trying to improve upon:
```
operation error SSO: GetRoleCredentials, https response error StatusCode: 403, RequestID: a41b0ac7-56bb-439c-acc6-6594d1532e62, api error ForbiddenException: No access
```

CC @jordiup

pkg/cfaws/assumer_aws_sso.go
```diff
@@ -141,6 +151,8 @@ func (c *Profile) SSOLogin(ctx context.Context, configOpts ConfigOpts) (aws.Cred
 func SSODeviceCodeFlow(ctx context.Context, cfg aws.Config, rootProfile *Profile) (*SSOToken, error) {
 	ssooidcClient := ssooidc.NewFromConfig(cfg)
 
+	shouldSilentStdout := ctx.Value("shouldSilentStdout")
```
Eddie023, line 154: I think there is a better way to pass value to context. Suggestions appreciated.

jordiup (approved):
Looks good @Eddie023, happy for this one to merge into `connecting-to-approvals`. We'll do another round of UAT 👌

PR #257, opened 2022-09-20
Add credential_process support, including the ability to request access to roles if CLI is used with Granted Approvals

jordiup:
### Description
1. Override native aws cli cmd to verify if users have the necessary roles and output redirect link to granted approvals if users don't have the required rules
2. Use `[credential_process](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html)` to source credentials  through granted cli. This PR extends functionality for the granted CLI, adding a `credential_process` compatible command that integrates with the native AWS CLI.

### Note 
User needs to update their `~./aws/config` profile to have a granted prefix. 
```
region = <>
granted_sso_account_id = <>
granted_sso_region = <>
granted_sso_role_name = <>
granted_sso_start_url = <>
credential_process = granted -- credentialsprocess --profile sandbox1 --url 
```

### Desired effect 
 Running `aws s3 ls --profile cf-dev` works via granted


pkg/cfaws/granted_config.go
```diff
@@ -0,0 +1,72 @@
...
+	cfg.SSOStartURL = gConfig.granted_sso_start_url
+	cfg.Region = gConfig.region
+
+	return &cfg, err
+}
+
+// For `granted login` cmd, we have to make sure 'granted' prefix
+// is added to the aws config file.
+func IsValidGrantedProfile(rawConfig configparser.Dict) error {
+	requiredGrantedCredentials := []string{"granted_sso_start_url", "granted_sso_region", "granted_sso_account_id", "granted_sso_role_name", "region"}
```
Eddie023, line 54: @chrnorm I have made the `region` field in `./aws/config` mandatory for granted prefixed configurations. However, I think `region` can be optional. Thoughts?

https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sso.html#:~:text=The%20presence%20of,the%20following%20example.

pkg/assume/assume.go
```diff
@@ -204,7 +205,19 @@ func AssumeCommand(c *cli.Context) error {
 
 	} else {
 		creds, err := profile.AssumeTerminal(c.Context, configOpts)
+		connectedToGrantedFlag := false
 		if err != nil {
+			serr, ok := err.(*smithy.OperationError)
+			if ok && connectedToGrantedFlag {
+				if serr.ServiceID == "SSO" {
+					baseUrl, ruleId := "granted.dev", "rul_2BtW97o6jTacUuzxNJZorACn5v0"
```
chrnorm, line 213: This hardcoded rule needs to be removed

pkg/cfaws/assumer_aws_sso.go
```diff
@@ -166,10 +175,12 @@ func SSODeviceCodeFlowFromStartUrl(ctx context.Context, cfg aws.Config, startUrl
 	}
 	// trigger OIDC login. open browser to login. close tab once login is done. press enter to continue
 	url := aws.ToString(deviceAuth.VerificationUriComplete)
-	fmt.Fprintf(color.Error, "If browser is not opened automatically, please open link:\n%v\n", url)
 
-	//check if sso browser path is set
+	if !shouldSilentLogs {
```
chrnorm, line 179: Do we need this? If we're printing to `color.Error` we're printing to stderr rather than stdout, so it shouldn't cause any issues?

pkg/cfaws/granted_config.go
```diff
@@ -0,0 +1,72 @@
...
+
+	for _, value := range requiredGrantedCredentials {
+		if _, ok := rawConfig[value]; !ok {
+			return fmt.Errorf("invalid aws config for granted login. %s is undefined but necessary", value)
```
chrnorm, line 58: Minor nitpick here to improve our error message

```suggestion
			return fmt.Errorf("invalid aws config for granted login. '%s' field must be provided", value)
```

pkg/config/config.go
```diff
@@ -25,6 +25,8 @@ type Config struct {
 	Keyring                *KeyringConfig `toml:",omitempty"`
 	Ordering               string
 	ExportCredentialSuffix string
+	// This is used to generate reference URls
+	GrantedApprovalsUrl string
```
chrnorm, line 29: nitpick: URL is an acronym and acronyms should be capitalised in Go

```suggestion
	GrantedApprovalsURL string
```

pkg/granted/credentialsprocess.go
```diff
@@ -0,0 +1,103 @@
...
+	Name:   "credentialsprocess",
+	Usage:  "",
+	Hidden: true,
+	Flags:  []cli.Flag{&cli.StringFlag{Name: "profile", Required: true}, &cli.StringFlag{Name: "url"}},
+	Action: func(c *cli.Context) error {
+
+		url := c.String("url")
+		profileName := c.String("profile")
+
+		if url == "" {
```
chrnorm, line 37: We should handle this better, as this assumes everyone who is using the `granted credentialsprocess` command is also using Granted Approvals. If the URL isn't provided, we should simply print an error message without a request URL.

pkg/granted/credentialsprocess.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		// In future we will support manual storing of url in config...
+		// conf, err := config.Load()
+		// if err != nil {
+		// 	log.Fatal("Failed to load Config for GrantedApprovalsUrl")
+		// }
+		// url = conf.GrantedApprovalsUrl
+		if url == "" {
+			log.Fatal("It looks like you haven't setup your GrantedApprovalsUrl\nTo do so please run: " + c.App.Name + " setup")
+		}
```
chrnorm, line 48: This can be removed as per the comment above - let's gracefully fall back if the URL isn't provided so that this command still works.

pkg/granted/credentialsprocess.go
```diff
@@ -0,0 +1,103 @@
...
+		profile, err := profiles.LoadInitialisedProfile(c.Context, profileName)
+		if err != nil {
+			_, ok := err.(*smithy.OperationError)
+			if ok {
+				// if it failed to load initialised try load it from the config
+				pr, err := profiles.Profile(profileName)
+				if err != nil {
+					log.Fatal(err)
+				}
+				log.Fatal(getGrantedApprovalsUrl(url, pr))
```
chrnorm, line 64: Does the output get printed if we return an error instead of calling `log.Fatal`? I would prefer to remove all calls to `log.Fatal`, as it prints an unnecessary timestamp before the message.

```suggestion
				return errors.New(getGrantedApprovalsUrl(url, pr))
```

If it doesn't work, can we replace this with:
```go
fmt.Println(getGrantedApprovalsUrl(url, pr))
os.Exit(1)
```

chrnorm: Note that this comment applies to all use of `log.Fatal` here as it's being used in a few spots

chrnorm (changes requested):
Looks good - my first pass of comments here based on a look over the code. I'm going to check the branch out and run some commands to test it out so might add some additional review comments based on the behaviour. Thanks @jordiup @Eddie023!

pkg/cfaws/assumer_aws_sso.go
Eddie023: Doesn't work. Yeah that was my understanding as well. I even tried to create a custom log and send to stderr but didn't work.

pkg/assume/assume.go
chrnorm: It looks like none of this is in use anyway as `connectedToGrantedFlag` is hardcoded to `false`? Can we remove all of this?

pkg/granted/credentialsprocess.go
jordiup: Good reccomendation @chrnorm. URL handling [amended here](https://github.com/common-fate/granted/commit/1cceea287a2f6aad32a820911d0b1bff8ca0ccaf), it now will only throw an error at the URL generation step (if still found to be empty string ""). This allows it to dual-function without an approvals deployment

pkg/granted/settings/request_url.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/AlecAivazis/survey/v2"
+	grantedConfig "github.com/common-fate/granted/pkg/config"
+	"github.com/pkg/errors"
+
+	"github.com/urfave/cli/v2"
+)
+
+var RequestURLCommand = cli.Command{
+	Name:  "request-url",
+	Usage: "Set the request url for credential_process command (connection to Grante dApprovals)",
```
lorengordon, line 18:
```suggestion
	Usage: "Set the request url for credential_process command (connection to Granted Approvals)",
```

chrnorm (approved):
Thanks @jordiup @Eddie023 this is looking good to go! And thankyou @lorengordon for helping catch the typo in the command usage text 😁

PR #258, opened 2022-09-20
Add a shortcut for Cost Explorer

mjulian:
This adds a shortcut for Cost Explorer (shortcut: `ce`).


meyerjrr (approved):
Thanks for the addition @mjulian!

PR #259, opened 2022-09-21
feat: Display credential duration in human-readable form

alexjurkiewicz:
The exact timestamp is hard to understand, even given it's in local time.


meyerjrr (commented):
Thanks for the QoL improvement here @alexjurkiewicz!  

I'm thinking that we might want to make this a configurable setting in the granted config, eg. `SimpleDurationMode` or something. Where the user is able to set whether they would like to view their expire as a duration (1hr) or timestamp for when it expires (2022-09-21 11:27:37 +0800 AWST)

Or event simpler, we just format it with both:
`session credentials will expire 2022-09-21 11:27:37 +0800 AWST (1 hour)`

Would love your thoughts on this @chrnorm @alexjurkiewicz

alexjurkiewicz:
Thanks for the reply. IMO

1. this behaviour is not important enough to be configurable -- you should
decide the best format and use that.
2. This string should be as simple as possible. It's not important enough
to take up much space. So, I favour my new format, then the original
format, and only finally your combined proposal.

chrnorm (approved):
Thanks for the contribution @alexjurkiewicz! And thanks for the review @meyerjrr.

In my view, a non-configurable simple duration is a good way to go here - I'm in favour of @alexjurkiewicz's change. I personally do not pay any attention to the current expiry time that `assume` prints, as it's a long string and is difficult to read. The expiry timestamp is exported as an env var too (`AWS_SESSION_EXPIRATION`) so if users wanted to do anything custom with it it's available in the shell.

It would be nice to avoid the additional external dependency, but Go's built-in duration parsing leaves a lot to be desired so I expect we'd just end up reimplementing most of `durafmt` ourselves if we didn't use it.

Issue #267, opened 2022-09-27
Granted 0.3.0 - invalid syntax


pjhavariotis:
I have just updated Granted to 0.3.0 and unfortunately, I cannot open a console to my existing AWS accounts.
My setup is Windows 10, with Firefox (latest version).
Below it the error I'm getting for one of my AWS accounts:
```
Opening a console for Production-Services in your browser...
parsing uid (S-1-5-21-273161372-3525547952-1395639491-311579): strconv.ParseUint: parsing "S-1-5-21-273161372-3525547952-1395639491-311579": invalid syntax
```

pjhavariotis:
I reverted back the update by installing the previous version 0.2.9.
The problem is gone!

chrnorm:
Thanks for the bug report @pjhavariotis! We dropped the [github.com/pkg/browser](https://github.com/pkg/browser) dependency as part of v0.3.0 which looks to have caused a couple of issues. We'll release a patch to fix these in Windows.

chrnorm:
@pjhavariotis we have released [v0.3.1](https://github.com/common-fate/granted/releases/tag/v0.3.1) which fixes this issue. Let us know if you experience any other issues!

pjhavariotis:
I have just tested it. Issue Fixed!

PR #269, opened 2022-09-28
Revert to using browser.OpenURL for SSO flow default

JoshuaWilkes:
This PR fixes issues with SSO flows on windows by reverting a recent change where we switched from using the browser package for launching the SSO login prompts.

It also fixes a bug with the forkprocess implementation for Windows.

This has been tested on Windows and Mac

### DIFF
--- a/pkg/forkprocess/forkprocess_windows.go
+++ b/pkg/forkprocess/forkprocess_windows.go
@@ -4,8 +4,6 @@ package forkprocess
 
 import (
 	"os/exec"
-	"os/user"
-	"strconv"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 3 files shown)


pkg/forkprocess/forkprocess_windows.go
```diff
@@ -20,22 +18,7 @@ type Process struct {
...
-		return nil, errors.Wrapf(err, "parsing uid (%s)", u.Uid)
-	}
-	gid, err := strconv.ParseUint(u.Gid, 10, 32)
-	if err != nil {
-		return nil, errors.Wrapf(err, "parsing gid (%s)", u.Uid)
-	}
-
 	p := Process{
-		UID:  uint32(uid),
-		GID:  uint32(gid),
```
meyerjrr, line 38: Was this stuff not needed?

JoshuaWilkes:
Fixes #267

meyerjrr (approved):

pkg/forkprocess/forkprocess_windows.go
JoshuaWilkes: No it was unused for windows because it just uses. This was causing one of the bugs in windows because these conversions were invalid

```
cmd := exec.Command(p.Args[0], p.Args[1:]...)
	err := cmd.Start()
```

PR #279, opened 2022-10-11
Added vpc and directconnect services

networkprogrammer:
Added VPC and Directconnect to the list of services as this is among the top go-to services for the networking folks.

### DIFF
--- a/pkg/console/service_map.go
+++ b/pkg/console/service_map.go
@@ -39,10 +39,13 @@ var ServiceMap = map[string]string{
 	"redshift":       "redshiftv2",
 	"sagemaker":      "sagemaker",
 	"ssm":            "systems-manager",
+	"vpc":            "vpc",
+	"dx":             "directconnect/v2"
 }
 
 var globalServiceMap = map[string]bool{
 	"iam":     true,
 	"route53": true,
 	"r53":     true,
+	"dx":	   true
 }


pkg/console/service_map.go
```diff
@@ -39,10 +39,13 @@ var ServiceMap = map[string]string{
 	"redshift":       "redshiftv2",
 	"sagemaker":      "sagemaker",
 	"ssm":            "systems-manager",
+	"vpc":            "vpc",
+	"dx":             "directconnect/v2"
```
chrnorm, line 43: Is `dx` a better abbreviation here than `dc`? I don't have much experience with Direct Connect so would be happy to accept your recommendation here @networkprogrammer.

chrnorm:
Thanks so much for the contribution @networkprogrammer and welcome to the community! I've left a quick question for you on the abbreviation for Direct Connect.

pkg/console/service_map.go
networkprogrammer: Hi @chrnorm , great question. If you were to ask 5 network folks, you are very likely to get 10 answers :)
DX would be the right abbreviation for directconnect.
All directconnect resources start with a dx, and a quick ctrl+f on the below page for directconnect, will show AWS using DX as an acronym

https://aws.amazon.com/directconnect/faqs/

chrnorm (approved):

PR #287, opened 2022-10-21
Fix tiny typo in help and error message

dserodio:
They show `~.aws/credentials` as the path to export credentials to, instead of `~/.aws/credentials`


chrnorm:
🙌 Thanks for fixing this up @dserodio!

chrnorm (approved):

PR #290, opened 2022-10-24
fix chrome profile resetting when '-r' is used

### LINKED ISSUE #289: Chrome sometimes forgets extensions in profiles
From a user in the [Common Fate Community Slack](https://join.slack.com/t/commonfatecommunity/shared_invite/zt-q4m96ypu-_gYlRWD3k5rIsaSsqP7QMg):

> I have switched to FF, and it's fine, but most of us use Chrome.  When granted launches a new instance of Chrome for a given AWS profile, and we configure it (e.g. with Okta and 1Password plugins, both of which are essential for our workflows) it remembers for some cases, then seems to forget and launch a new bare Chrome.

It appears that running `assume` with a region, i.e.

```
assume -c dev -r us-west-2
```

seems to reset the Chrome profile.

chrnorm:
Fixes #289. I can't tell *why* it works though.

Previously we hashed the profile names when setting up a Chrome profile folder.

You'd end up with a file structure looking like this:

```
~/.granted
├── aws_profiles_frecency
├── chromium-profiles
│   └── 1
│       ├── 2731494339
│       ├── AutofillStates
│       ├── CertificateRevocation
│       ├── ClientSidePhishing
│       ├── Crowd Deny
│       ├── DesktopSharingHub
│       ├── FileTypePolicies
│       ├── FirstPartySetsPreloaded
│       ├── GrShaderCache
│       ├── Last Version
│       ├── Local State
│       ├── MEIPreload
│       ├── OnDeviceHeadSuggestModel
│       ├── OptimizationHints
│       ├── OriginTrials
│       ├── PKIMetadata
│       ├── RecoveryImproved
│       ├── SSLErrorAssistant
│       ├── Safe Browsing
│       ├── SafetyTips
│       ├── ShaderCache
│       ├── Subresource Filter
│       ├── System Profile
│       ├── UrlParamClassifications
│       ├── Variations
│       ├── Webstore Downloads
│       ├── WidevineCdm
│       ├── ZxcvbnData
│       ├── chrome_shutdown_ms.txt
│       └── persisted_first_party_sets.json
└── config
```

Where `2731494339` is a Chrome profile:

```
~/.granted/chromium-profiles/1/2731494339/
├── AutofillStrikeDatabase
├── BudgetDatabase
├── Cache
├── Code Cache
├── Download Service
├── Extension Rules
├── Extensions
├── Feature Engagement Tracker
├── GCM Store
├── IndexedDB
├── LOCK
├── LOG
├── Network Action Predictor
├── Network Action Predictor-journal
├── Network Persistent State
├── Preferences
├── Search Logos
├── Secure Preferences
├── Segmentation Platform
├── Service Worker
├── Sessions
├── Shortcuts
├── Shortcuts-journal
├── Storage
├── TransportSecurity
├── WebStorage
├── databases
├── optimization_guide_hint_cache_store
├── optimization_guide_model_metadata_store
└── optimization_guide_prediction_model_downloads

20 directories, 10 files
```

This PR switches the profile folder to be `granted-profile-<PROFILE_NAME>`. This is easier to understand and also seems to fix the problem - when running a dev build I was able to swap between regions without the Chrome extensions being corrupted.

I would have expected the hash to be different, based on the `-r` flag, but it is the same each time:

```
❯ dassume -c demo-sandbox1

ℹ️  use -s to open a specific service (https://docs.commonfate.io/granted/usage/console)

Opening a console for demo-sandbox1 in your browser...
hash: 2731494339

❯ dassume -c demo-sandbox1 -r us-west-2

ℹ️  use -s to open a specific service (https://docs.commonfate.io/granted/usage/console)

Opening a console for demo-sandbox1 in your browser...
hash: 2731494339
```

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -229,22 +229,22 @@ func AssumeCommand(c *cli.Context) error {
 		case browser.ChromeKey:
 			l = launcher.ChromeProfile{
 				ExecutablePath: browserPath,
-				UserDataPath:   path.Join(grantedFolder, "chromium-profiles", "1"), // held over for backwards compatibility, "1" indicates Chrome profiles
+				UserDataPath:   path.Join(grantedFolder, "chromium-profiles", "chrome"),
 			}
 		case browser.BraveKey:
 			l = launcher.ChromeProfile{
 				ExecutablePath: browserPath,
-				UserDataPath:   path.Join(grantedFolder, "chromium-profiles", "2"), // held over for backwards compatibility, "2" indicates Brave profiles
+				UserDataPath:   path.Join(grantedFolder, "chromium-profiles", "brave"),
 			}
 		case browser.EdgeKey:
 			l = launcher.ChromeProfile{
 				ExecutablePath: browserPath,
-				UserDataPath:   path.Join(grantedFolder, "chromium-profiles", "3"), // held over for backwards compatibility, "3" indicates Edge profiles
+				UserDataPath:   path.Join(grantedFolder, "chromium-profiles", "edge"),
 			}
 		case browser.ChromiumKey:
 			l = launcher.ChromeProfile{
 				ExecutablePath: browserPath,
-				UserDataPath:   path.Join(grantedFolder, "chromium-profiles", "4"), // held over for backwards compatibility, "4" indicates Chromium profiles
+				UserDataPath:   path.Join(grantedFolder, "chromium-profiles", "chromium"),
 			}
 		case browser.FirefoxKey:
 			l = launcher.Firefox{
--- a/pkg/launcher/chrome_profile.go
+++ b/pkg/launcher/chrome_profile.go
@@ -1,8 +1,7 @@
 package launcher
 
 import (
-	"fmt"
-	"hash/fnv"
+	"strings"
 )
 
 type ChromeProfile struct {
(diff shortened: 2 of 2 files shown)


tomharrisonjr:
Happy to run a test build locally if you want confirmation.  Thanks @chrnorm !!

chrnorm:
Note that this is a breaking change - when merged/released this will cause users to lose any customisations they have made to their Chrome profiles. Depending on community feedback we may need to provide a migration script for this. Given that this is a breaking change I have also taken the opportunity to give the different browser folders (`1`, `2`, `3`, `4`) human-readable names (`chrome`, `brave`, `edge`, `chromium`)

chrnorm:
Yes please @tomharrisonjr! If you'd like to build locally, you can clone this repo and run
```
make cli
```

The CLI will then be available as `dgranted` and `dassume`.

JoshuaWilkes:
I'm sure there was some reason we used a hash but there is no comment there 😓

JoshuaWilkes:
Digging into this, we had previously included the region in profile names, then reverted this because you could have multiple regions open in different tabs, however we only reverted it for firefox at the time.

Then @chrnorm refactored the profile name method and removed the region from the hash.
This was the original method
```
func (r *BrowserOpts) MakeExternalProfileTitle() string {
	n := r.Profile
	if r.Region != "" {
		n = r.Profile + "(" + r.Region + ")"

	}

	h := fnv.New32a()
	h.Write([]byte(n))

	hash := fmt.Sprint(h.Sum32())
	return hash

}
```

So overall I'm happy to remove the hash all together so long as we correctly escape any characters which are valid in aws profile names but not in file names.

Invalid characters for folder names https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names

invalid character for AWS profile names illegalChars := "\\][;'\""

We could probably regex replace all non alphanumeric characters with - I think the following snippet would work

```
var nonAlphanumericRegex = regexp.MustCompile(`[^a-zA-Z0-9]+`)

func clearString(str string) string {
    return nonAlphanumericRegex.ReplaceAllString(str, "-")
}
```

chrnorm:
Thanks @JoshuaWilkes, let me take a quick pass over this to edit. The format I will test is:

```
PROFILE_NAME-HASH
```

I will use the regex you have suggested. I am still not sure why we are seeing the extensions being corrupted if FNV is used by itself. As you mentioned, we're no longer using regions as part of the data to be hashed, so the hash itself isn't changing.

chrnorm:
Closing - stale

PR #293, opened 2022-10-26
CF-188 Add profile registry feature

Eddie023:
RFD: https://github.com/common-fate/rfds/discussions/2

Docs: https://github.com/common-fate/docs/pull/136

Summary:
1. Add "granted registry" command to add/remove/sync aws config registries from git repository to aws config file.


pkg/granted/registry/add.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	for _, file := range dir {
+		if file.Name() == "granted.yml" || file.Name() == "granted.yaml" {
+			return nil, true
+		}
+	}
+
+	return nil, false
+}
```
chrnorm, line 94: minor nitpick here - you just make this return an error and it will be simpler

```suggestion
func isValidRegistry(folderpath string, url string) error {
	dir, err := os.ReadDir(folderpath)
	if err != nil {
		return err
	}

	for _, file := range dir {
		if file.Name() == "granted.yml" || file.Name() == "granted.yaml" {
			return nil
		}
	}

	return fmt.Errorf("unable to find `granted.yml` file in %s", url)
}
```

pkg/granted/registry/add.go
```diff
@@ -0,0 +1,94 @@
...
+	"strings"
+
+	"github.com/urfave/cli/v2"
+)
+
+var AddCommand = cli.Command{
+	Name: "add",
+	Action: func(c *cli.Context) error {
+
+		if c.Args().Len() != 1 {
```
chrnorm, line 18: Make sure to test with SSH as well as HTTPS - you should still be able to use the `net/url` library to parse these. e.g. `git@github.com:common-fate/granted-registry.git`

pkg/granted/registry/add.go
```diff
@@ -0,0 +1,197 @@
...
+// we use this to allow flags to be set on either side of the profile arg e.g `assume -c profile-name -r ap-southeast-2`
+func GlobalFlags() []cli.Flag {
+	return []cli.Flag{
+		&cli.StringFlag{Name: "ref", Aliases: []string{"r"}, Usage: "Used to reference a specific commit hash, tag name or branch name"},
+	}
+}
+
+var AddCommand = cli.Command{
+	Name:  "add",
+	Flags: GlobalFlags(),
```
chrnorm, line 28: It is simpler to just put the flags directly here on the command (it makes it a lot easier to see what flags relate to a particular command, rather than needing to look up what `GlobalFlags` is)

```suggestion
	Flags: []cli.Flag{
 		&cli.StringFlag{Name: "ref", Aliases: []string{"r"}, Usage: "Used to reference a specific commit hash, tag name or branch name"},
 	}
```

pkg/granted/registry/add.go
```diff
@@ -0,0 +1,197 @@
...
+	Action: func(c *cli.Context) error {
+
+		addFlags, err := cfflags.New("assumeFlags", GlobalFlags(), c, 3)
```
chrnorm, line 31: Rather than using `cfflags` here we can just use the CLI package's built in flag parsing. This should save the need for the 'depth' argument that you've added (more info on this in comments below)

pkg/granted/registry/add.go
```diff
@@ -0,0 +1,197 @@
...
+
+					}
+				}
+
+			} else {
+				return err
+			}
+		} else {
+			//if a specific ref is passed we will checkout that ref
+			if addFlags.String("ref") != "" {
```
chrnorm, line 110: You can just use the built-in parsing here

```suggestion
			if c.String("ref") != "" {
```

pkg/urfav_overrides/flags.go
```diff
@@ -24,9 +24,13 @@ type Flags struct {
...
 //
-// allFlags := cfflags.New("name",GlobalFlagsList, c)
+// allFlags := cfflags.New("name",GlobalFlagsList, c, 1)
 // allFlags.String("region")
-func New(name string, flags []cli.Flag, c *cli.Context) (*Flags, error) {
+// When using granted global flags for a subcommand. Eg. granted registry add {URL:} -c ref. you will need to need to specify a command depth
+// for it to work correctly.
+// The depth is used to filter out the subcommands when grabbing out the flags passed in the body of the cli request
+// err = set.Parse(ag) fails to populate the flag value when args includes subcommands so we filter them out
+func New(name string, flags []cli.Flag, c *cli.Context, depth int) (*Flags, error) {
```
chrnorm, line 33: The purpose of this package is to allow flags to be provided in any order for the `assume` command.

e.g.

```
assume -c my-profile
assume my-profile -c
```
The built in `urfav/cli` argument parser doesn't support this so we needed to extend it ourselves.

For the `granted registry` commands though we don't need to support flags being placed in any order.
For the `--ref` flag as an example, the only accepted order should be:

```
granted registry add --ref dev <registry url>
```

This behaviour is how `urfav/cli` works already, so we don't need to use this package at all.

Looking at the changes in this PR it looks like you have attempted to extend the `urfav_overrides` package to get this behaviour, but it is already built in.

I would propose that we revert the changes on this package (i.e. remove the `depth` argument), and then just use built in argument parsing, i.e.

```go
var MyCommand = cli.Command{
	Name: "command",
	Flags: []cli.Flag{
		&cli.StringFlag{Name: "ref"},
	},
	Action: func(c *cli.Context) error {
		ref := c.String("ref")
		// do stuff with ref

		return nil
	},
}
```

pkg/granted/registry/template.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"bytes"
+	"text/template"
+)
+
+type Template struct {
+	RepoURL string
+}
+
+const AUTO_GENERATED_MSG string = `### GRANTED-REGISTRY-SECTION: "{{.RepoURL}}". DO NOT EDIT.
```
chrnorm, line 12: nitpick: can you make this match the autogenerated description in the RFD? This is very very minor but I think that the uppercase is less aesthetically nice than `# granted-start: https://github.com/octo-org/granted-registry.git`

https://github.com/common-fate/rfds/discussions/2

```suggestion
const AUTO_GENERATED_MSG string = `### GRANTED-REGISTRY-SECTION: "{{.RepoURL}}". DO NOT EDIT.
```

pkg/config/config.go
```diff
@@ -28,7 +28,8 @@ type Config struct {
 	// AccessRequestURL is a Granted Approvals URL that users can visit
 	// to request access, in the event that we receive a ForbiddenException
 	// denying access to assume a particular role.
-	AccessRequestURL string `toml:",omitempty"`
+	AccessRequestURL   string `toml:",omitempty"`
+	ProfileRegistryURL string `toml:",omitempty"`
```
chrnorm, line 32: We should change this to a type that can support multiple registries - we should also store the `ref` for each registry.

chrnorm (changes requested):
Looking good @Eddie023, have added some feedback on flag parsing

pkg/urfav_overrides/flags.go
Eddie023: @meyerjrr Tagging you here as this was implemented by you.

pkg/config/config.go
Eddie023: This has been done in my current changes. 🎉

Eddie023: `ref` implementation is still WIP.

pkg/granted/registry/template.go
Eddie023: @chrnorm We might need to remove `repoURL` from the autogenerated description cause we can have multiple repos now.

pkg/granted/registry/add.go
Eddie023: Was not able to use this for SSH. I have reverted to writing a custom regex parser for now. Resolving this for now.

pkg/urfav_overrides/flags.go
chrnorm: @Eddie023 just revert these changes directly rather than waiting for Jack's okay!

pkg/granted/registry/remove.go
```diff
@@ -0,0 +1,77 @@
...
+	Usage:       "Remove subscribing to existing registry",
+	Action: func(c *cli.Context) error {
+
+		gConf, err := grantedConfig.Load()
+		if err != nil {
+			return err
+		}
+
+		if len(gConf.ProfileRegistryURLS) <= 0 {
+			clio.Error("There are no granted registry configured currently.\n Please use 'granted registry add <https://github.com/your-org/your-registry.git>' to add a new git repository")
```
chrnorm, line 26:
```suggestion
			clio.Error("There are no profile registries configured currently.\n Please use 'granted registry add <https://github.com/your-org/your-registry.git>' to add a new git repository")
```

pkg/granted/registry/registry_cmd.go
```diff
@@ -0,0 +1,37 @@
...
+	Subcommands: []*cli.Command{&AddCommand, &SyncCommand, &RemoveCommand},
+	Action: func(c *cli.Context) error {
+
+		gConf, err := grantedConfig.Load()
+		if err != nil {
+			return err
+		}
+
+		if len(gConf.ProfileRegistryURLS) <= 0 {
+			clio.Info("There are no registry subscribed to granted yet.\n Use 'granted registry add <your_repo>' to configure subscription.\n")
```
meyerjrr, line 21: Reading over the prompts here, should we switch the wording from: `registry` to `repository` since we are referring to git repos?

```
There are no repositories subscribed to granted yet.\n Use 'granted registry add <your_repo>' to configure subscription.
```

JoshuaWilkes (approved):

JoshuaWilkes:
For profile descriptions if any profiles have a description, then headers are added to the assume prompt, if no profiles have descriptions, the header is omitted

image

JoshuaWilkes (approved):
Great Work!

PR #297, opened 2022-10-28
Use ini package for config file parsing

JoshuaWilkes:
This PR swaps out configparser in favour of gopkg.in/ini.v1

The swap is mostly like for like and I have not made an attempt at refactoring our parsing logic in this PR.

The benefit gained from this swap is that comments are now supported in configuration files. 

When performing write operations, Granted will no longer remove comments from the configuration files!

The ini package makes it simple to write tests where we define the config files using a string rather than reading and writing to the filesystem


pkg/cfaws/granted_config.go
```diff
@@ -5,68 +5,54 @@ import (
...
-			return &config.SharedConfig{}, nil
-		}
 		return nil, err
 	}
-
-	cfg.SSOAccountID = gConfig.granted_sso_account_id
-	cfg.SSORegion = gConfig.granted_sso_region
-	cfg.SSORoleName = gConfig.granted_sso_role_name
-	cfg.SSOStartURL = gConfig.granted_sso_start_url
-	cfg.Region = gConfig.region
```
Eddie023, line 46: @JoshuaWilkes `cfg.Region` is an optional field but we should still populate the`cfg.Region` field if user has it available in "aws config" file. I remember some assume commands failing because `region` was not set.

pkg/cfaws/granted_config.go
```diff
@@ -5,68 +5,54 @@ import (
...
-func (gConfig *grantedSSOConfig) ConvertToAWSConfig(ctx context.Context, p *Profile) (*config.SharedConfig, error) {
-	cfg, err := config.LoadSharedConfigProfile(ctx, p.Name, func(lsco *config.LoadSharedConfigOptions) { lsco.ConfigFiles = []string{p.File} })
-	// if required profile doesn't exist then
-	// return empty config instead of error.
+	cfg, err := config.LoadSharedConfigProfile(ctx, profile.Name, func(lsco *config.LoadSharedConfigOptions) { lsco.ConfigFiles = []string{profile.File} })
 	if err != nil {
-		if _, ok := err.(config.SharedConfigProfileNotExistError); ok {
-			return &config.SharedConfig{}, nil
-		}
 		return nil, err
```
Eddie023, line 18: Also, `config.LoadSharedConfigProfile` method just hangs if this method is called with the profile that has "granted_" prefix. `IsValidGrantedProfile` check should catch that but just letting you know. The reason I returned empty config instead of 'err' was due to this reason.

pkg/cfaws/granted_config.go
```diff
@@ -5,68 +5,54 @@ import (
...
+	item, err = profile.RawConfig.GetKey("granted_sso_role_name")
+	if err != nil {
+		return nil, err
+	}
+	cfg.SSORoleName = item.Value()
+	item, err = profile.RawConfig.GetKey("granted_sso_start_url")
+	if err != nil {
+		return nil, err
+	}
+	cfg.SSOStartURL = item.Value()
```
Eddie023, line 39: Not sure if we can replace this with a loop 
```suggestion
          requiredGConfig = ["a", "b", "c", "d"]
          
          for _, key := range requiredGConfig {
          item, err := profile.RawConfig.GetKey(key)
	if err != nil {
		return nil, err
	}
	cfg[key] = item.Value()
          }
```

pkg/cfaws/granted_config.go
JoshuaWilkes: This works as expected

pkg/cfaws/granted_config.go
JoshuaWilkes: This still works as expected

pkg/cfaws/granted_config.go
JoshuaWilkes: I think this is about as simple as it gets

Eddie023 (approved):

Issue #310, opened 2022-11-18
Granted Registries do not support Gerrit repositories


sosheskaz:
Granted parses the repository string to look for organization name, but not all git servers have organization names, such as [Gerrit](https://www.gerritcodereview.com/).

Gerrit repositories generally exist in the format `https://gerrit.googlesource.com/gerrit`. This reliably produces the following error:

```bash
$ granted --verbose registry add https://gerrit.googlesource.com/gerrit
[DEBUG] parsing the provided url to get host, organization and repo name for https://gerrit.googlesource.com/gerrit
[✘] unable to parse the provided git url 'https://gerrit.googlesource.com/gerrit'
```

sosheskaz:
Some git webservers may also have unexpected formats, ex. if they are running on a server that does path-pased forwarding, ex: `https://my-website.com/gitea/organization/repo`. I do not know how well it would handle that case.

JoshuaWilkes:
@sosheskaz I have continued the conversation on https://github.com/common-fate/rfds/discussions/2 with a proposal for how we can improve the support for non Github URLs, if you have any thoughts or feedback on the proposal please feel free to join the discussion there.

Eddie023:
Should be fixed with #317. Closing the issue.

PR #317, opened 2022-12-01
CLI-88 Update Profile Registry Implementation 

Eddie023:
# Summary

[[Community Member raised an issue where he couldn’t use Gerrit Repository for Profile Registry.](https://github.com/common-fate/granted/issues/310)](https://github.com/common-fate/granted/issues/310) 

This led to some thinking in terms of design decisions and [[we opted to update the implementation](https://github.com/common-fate/rfds/discussions/2#discussioncomment-4192655)](https://github.com/common-fate/rfds/discussions/2#discussioncomment-4192655) to remove some restriction and additionally add other required features asked by community members. 

# Breaking Change

This is a breaking change. If users were already using Profile Registry then they can use the `granted registry migrate` command to update their configuration to the new data structure. 

Users will be warned about this in `autosync` if they were already configured with a Profile Registry. 

# Docs PR
https://github.com/common-fate/docs/pull/174

# Changes Made on Implementation

### 1. Removal of Regex to Parse Git URLs

→ Instead of using custom regex to parse `organization` `hostname` `repository name` to figure out the folder name on which we should clone the repo and if the URL is valid. Just `git clone` the repo and output the error if any. 

→ For figuring out folder name, a new `-name` flag is added. More on this is below. 

### 2. Data structure is changed from Array of git-urls to ProfileRegistry Struct
This gives us more flexibility in adding global as well as registry-level configs.

### 3. Passing information as flags instead of single string to granted registry add cmd

Previously, every piece of information was passed as an argument to `granted registry add` as 

```bash
> granted registry add git@github.com:Eddie023/repo.git/subfolder/devConfig.yml
```

Now, multi flags as defined such as 

1. `--name`: Name of the registry. This is a Required field and needs to be unique. Used to create folder name when cloning the repository. 

 2.   `--url`: We use `git clone <value>` and let git throw an error if any. This is also a Required field.

1.  `--path`: Optional field. Useful when only a subfolder needs to be synced. 
2. `--filename`: Optional field. Can be used when YAML config file name is not `granted.yml`
3. `--priority`: [[1. Deterministic Priority ](https://www.notion.so/1-Deterministic-Priority-8c5875ab69174a31b619f1100d582f9f)](https://www.notion.so/1-Deterministic-Priority-8c5875ab69174a31b619f1100d582f9f) 
4. `--requiredKeys` : Can be used to bypass the prompt for required keys for the `granted registry add` cmd 
5. `prefix-duplicate-profiles`: Can be used to bypass the prompt to prefix duplicate profiles 
6. `prefix-all-profiles`: Can be used to add prefixes to specific registry 

### 4. Updates to prefixing registry names to duplicate profiles

Instead of Automatically prefixing names to duplicate profiles, now users will be prompted if they want to update profile names or if they want to abort sync and fix it manually. 

# New Features

## 1. Deterministic Priority

A community member mentioned that there should be consistency when prefixing registry name to duplicate profiles. For example: 

```bash
[granted_registry_start A] ## priority 0 ## default
[Profile dev] 

[granted registry_start B] ## priority 5 
[Profile dev]
```

Now, with the priority flag, the registry with the highest priority will have the default profile name, and the registry with lesser priority will get the prefix on duplicate profiles. 

```bash
[granted_registry_start B]
[Profile dev] 

[granted registry_start A] 
[Profile A.dev]
```

## 2. Ability to template AWS profiles

Community Users wanted the ability to template AWS profiles for reusability or for the addition of user-specific values. 

Users can add any arbitrary keys to the `granted.yml` file inside `templateValues` as:

```bash
awsConfig:
  - ./team_a/config1
  - ./team_b/config2
templateValues:
  - ReadOnlyBaseProfile:
      - value: "my-profile"
  - SessionName: 
      - isRequired: true 
      - prompt: "Enter your user name (e.g john.doe)" 
```

### 2.1 User-Specific Value

In cases as shown below, admin would want some user specific values to be entered. 

```bash
[profile complex]
granted_sso_start_url = https://abc.awsapps.com/start
credential_process = granted credential-process --profile {{ .Profile }} --url https://internal.prod.granted.run
session_name = {{ .Required.SessionName }}
```

In this case, when the user runs `granted registry add/sync`  they will be prompted to add the required `SessionName` key before the registry is synced. Note, `isRequired:true` should be present for User Specific Values to work

### 2.2 **Arbitrary registry-level variables**

In case, where admin wants to reuse the AWS profile values, they can do so by adding the `-value` key to any variables. 

```bash
[profile myacct-ro]
source_profile = {{ .Variables.ReadOnlyBaseProfile }}
role_arn = arn:aws:iam::0123456789012:role/readonly
```


JoshuaWilkes (commented):
Requesting that we rename update to migrate

JoshuaWilkes (approved):

lorengordon:
@Eddie023 @JoshuaWilkes Did you mean to create tag 0.16.0 after merging this PR?

Issue #183, opened 2022-07-10
[Feature Request] SSO Browser Profile


RichiCoder1:
For browsers that support profiles (Chrome, Edge, and Chromium Variants), it'd be great to not only launch a specific browser for SSO but to also support launching a specific browser _profile_.

For context, I have a separate Work profile already which is ideally the one launched for SSO sessions.

john-fitsense:
This would be useful for people, such as myself, who use multiple browser profiles (eg work vs personal).
Is there a way to invoke the browser with a specific profile (argument)?

It looks like the pull request was rejected/abandoned!?  I can't find any of the solution in master?  Is this correct?  So the issue remains open.

johnmee:
This would be useful for people, such as myself, who use multiple browser profiles (eg work vs personal).
Is there a way to invoke the browser with a specific profile (argument)?

It looks like the pull request was rejected/abandoned!? I can't find any of the solution in master? Is this correct? So the issue remains open.

shwethaumashanker:
With [v0.16.0](https://github.com/common-fate/granted/releases/tag/v0.16.0), Launching a specific browser profile is now possible with `--browser-profile`. Try `assume profile-name -c --browser-profile "<your_browser_profile>"`

johnmee:
No joy: still opens the default profile:
* Firefox v117.0.1 (64-bit) snap for ubuntu
* Granted Containers v1.1.2 on Firefox installed to both default and alternate profiles

`firefox -p AltProfile` successfully opens firefox with the alternate profile. Their configs are held under `/home/<username>/snap/firefox/common/.cache/mozilla/firefox/<randomstring>.<profilename>`

`assume <awsprofilename> -c --browser-profile AltProfile`  still opens up firefox with the default browser profile.  I tried putting AltProfile in double quotes or using the registered email address for the profile to no avail.

Not sure what I'm doing wrong :-(

meyerjrr:
Hey @johnmee, this fix was specifically for chromium profiles! But we can definitely expand this out for Firefox profiles too.
I've made a new issue that we can use to track progress on: #486

PR #318, opened 2022-12-01
Added logic for adding custom chrome profiles

meyerjrr:
Adds ability to specify a chromium profile name to open up into


meyerjrr:
Fixes #183

pkg/launcher/chrome_profile.go
```diff
@@ -12,24 +17,140 @@ type ChromeProfile struct {
...
+
+// 	h := fnv.New32a()
+// 	h.Write([]byte(profile))
+
+// 	hash := fmt.Sprint(h.Sum32())
+// 	return hash
+// }
+
+// todo: verify these are correct
+var BravePathMac = "Library/Application Support/BraveSoftware/Brave-Browser/Local State"
```
Eddie023, line 48: Looks like a bit of cleaning is required around here 😅

pkg/assume/entrypoint.go
```diff
@@ -38,6 +38,7 @@ func GlobalFlags() []cli.Flag {
 		&cli.StringFlag{Name: "sso-region", Usage: "Use this in conjunction with --sso, the sso-region"},
 		&cli.StringFlag{Name: "account-id", Usage: "Use this in conjunction with --sso, the account-id"},
 		&cli.StringFlag{Name: "role-name", Usage: "Use this in conjunction with --sso, the role-name"},
+		&cli.StringFlag{Name: "profile", Usage: "Use a pre-existing profile in your browser"},
```
JoshuaWilkes, line 41: lets call this browser profile to avoid confusion with `aws --profile myprofile`
```suggestion
		&cli.StringFlag{Name: "browser-profile", Usage: "Use a pre-existing profile in your browser"},
```

pkg/launcher/chrome_profile.go
meyerjrr: @JoshuaWilkes when you verified all the locations on each OS did you do this manually or through Googling each location. Can only verify OSX until next year when I have access to a Windows + linux PC

JoshuaWilkes: @meyerjrr I was able to verify all of them manually, as I have access to all 3

meyerjrr:
Reopening due to interest on #183 

Keeping in mind that due to the nature of how profiles work. Being able to open up into a specific profile will break the multi region functionality of granted.

- This is due to profile names being the differentiator when opening up the console for different regions.

Eddie023 (approved):

PR #320, opened 2022-12-15
Use slash to separate account and role name in `sso generate`

dserodio:
Currently a dash is used both in the normalized AWS Account Name and to separate it from the role name.

Using a slash to separate the account and role names makes the generated profile names easier to parse, both visually and also for customizing the generated file using a regexp.

Currently:
```[profile Long-AWS-Account-Name-Long-Role-Name]```

After this PR:
```[profile Long-AWS-Account-Name/Long-Role-Name]```


JoshuaWilkes (approved):
Happy to go with this, in future we may want to make this more configurable

PR #321, opened 2022-12-21
Better logging around waiting for the browser

grantjoy:
It can take a long time for the browser authentication piece to finish, and some users get antsy and will kill the process. This PR attempts to give them some confidence that things are still happening in the background and that they should be patient.

### DIFF
--- a/pkg/cfaws/assumer_aws_sso.go
+++ b/pkg/cfaws/assumer_aws_sso.go
@@ -183,8 +183,7 @@ func SSODeviceCodeFlowFromStartUrl(ctx context.Context, cfg aws.Config, startUrl
 	}
 	// trigger OIDC login. open browser to login. close tab once login is done. press enter to continue
 	url := aws.ToString(deviceAuth.VerificationUriComplete)
-	clio.Info("If the browser does not open automatically, please open this link:")
-	clio.Info(url)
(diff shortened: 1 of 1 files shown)


pkg/cfaws/assumer_aws_sso.go
```diff
@@ -214,7 +213,7 @@ func SSODeviceCodeFlowFromStartUrl(ctx context.Context, cfg aws.Config, startUrl
 		}
 	}
 
-	clio.Info("Awaiting authentication in the browser...")
+	clio.Info("Awaiting browser authentication, this may take up to 2 minutes...")
```
JoshuaWilkes, line 216: Thanks @grantjoy 
What do you think about describing the steps a little more?

```suggestion
	clio.Info("Awaiting AWS authentication in the browser")
	clio.Info("You will be prompted to authenticate with AWS in the browser, then you will be prompted to 'Allow'")
	clio.Info("Once you press 'Allow', the sign in process will complete in a few seconds")
```

grantjoy: Great idea! Committed.

grantjoy (commented):
cool

JoshuaWilkes (approved):

PR #327, opened 2022-12-28
generate and populate profile name with custom go template

vdesjardins:
Add an additional generate and populate parameter to use a go template to customize the profile name.


JoshuaWilkes:
Thanks for the contribution @vdesjardins looks great.
I was considering that we needed to be able to customise this further for some users and this covers that well.

Would you be able to make a PR into the docs repo with some example usage of the templates flag?

You can add a section to the bottom of this page with a few examples of custom templates.
https://github.com/common-fate/docs/blob/main/docs/granted/usage/automatic-config-generation.md

I haven't come across the template functions package before however it looks good so happy to include it here, we can consider it's usage for other features like profile registries as well.

JoshuaWilkes (changes requested):
One thing I am concerned about is that there is no validation on the profile names or the template.

This is the set of invalid characters for a section name 

```
illegalChars := ` \][;'"`
```

It would be great if parseCliOptions function included a validation that the profile name template does not include the characters so it's harder to misuse this feature.

The same goes for the --prefix flag

pkg/granted/sso.go
```diff
@@ -245,7 +265,3 @@ func mergeSSOProfiles(config *ini.File, prefix string, ssoProfiles []SSOProfile)
 
 	return nil
 }
-
-func normalizeAccountName(accountName string) string {
```
JoshuaWilkes, line 249: Did you have a specific use-case in mind where you want the account name normalised in a different way?

pkg/granted/sso.go
```diff
@@ -16,6 +18,8 @@ import (
 	"gopkg.in/ini.v1"
 )
 
+var defaultProfileNameTemplate = "{{.AccountName | replace \" \" \"-\"}}/{{.RoleName}}"
```
JoshuaWilkes, line 21: To me, normalisation of the .AccountName feels like something that should be done outside the template. So a user is able to write a template like so.

If accountName is already normalised in the available template data then we can drop the requirement for github.com/Masterminds/sprig/v3 unless you can think of more complex use-cases someone may want to use?

```go
type SSOProfile struct {
	// SSO details
	StartUrl  string
	SSORegion string
	// Account and role details
	AccountId   string
	AccountName string
	RoleName    string
}

nameTemplate := "{{ .AccountName }}/{{ .RoleName }}"

or

nameTemplate := "{{ .AccountName }}_{{ .RoleName }}"
```

vdesjardins: I agree that normalization can happen before template is applied. For sprig I added it mainly for the replace but it gives many functions to manipulate text that gives more freedom to generate profile name. Personally I don't need it. Happy to remove that dependency from this PR.

pkg/granted/sso.go
```diff
@@ -94,39 +99,41 @@ var PopulateCommand = cli.Command{
...
 	}
 
 	ssoRegion, err := cfaws.ExpandRegion(c.String("region"))
 	if err != nil {
 		return nil, err
 	}
 
+	profileTemplate := c.String("profile-template")
+	if strings.ContainsAny(profileTemplate, profileSectionIllegalChars) {
+		return nil, fmt.Errorf("--profile-template flag must not contains illegal characters (%s)", profileSectionIllegalChars)
```
JoshuaWilkes, line 113:
```suggestion
		return nil, fmt.Errorf("--profile-template flag must not contain any of these illegal characters (%s)", profileSectionIllegalChars)
```

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):
Thanks for making the requested changes, this looks good 👍

Issue #338, opened 2023-01-18
Profile Registry should sync config files that has 'sso-session' section 


Eddie023:
Currently, each non-default section is checked to see if it is a valid AWS profile name and skipped if not.
https://github.com/common-fate/granted/blob/main/pkg/granted/registry/sections.go#L187.

PR #339, opened 2023-01-18
copy any none profile section from config file in Profile Registry

Eddie023:
## Describe your changes
While syncing Profiles in Profile Registry, we were checking if the passed profiles are valid AWS profiles This was troublesome because now you cannot add `[sso-session abc]` section in your config file. This change will sync any content that is not an AWS profile as it is. 

Will only warn if the AWS config with `profile` prefix have any invalid characters in it.  

- Fixes: https://github.com/common-fate/granted/issues/338
 
## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Issue and Documentation
 https://github.com/common-fate/granted/issues/338

## Testing

1. Add any section without `[profile abc]` in your remote registry. 
2. run `granted registry add/sync` cmd 
3.  Should see the content in your local aws config file.

- [x] I have added tests that prove my fix is effective or that my feature works.
- [x] New and existing unit tests pass with my changes.
- [x] Any dependent changes have been merged and published in downstream modules.

### DIFF
--- a/pkg/granted/registry/sections.go
+++ b/pkg/granted/registry/sections.go
@@ -153,16 +153,11 @@ func getNonGrantedProfiles(config *ini.File) []*ini.Section {
 	return nonGrantedProfiles
 }
 
-func generateNewRegistrySection(r *Registry, configFile *ini.File, clonedFile *ini.File, opts syncOpts) error {
+func generateNewRegistrySection(r *Registry, configFile *ini.File, clonedFile *ini.File, gconf *grantedConfig.Config, opts syncOpts) error {
 	sectionName := r.Config.Name
 	clio.Debugf("generating section %s", sectionName)
 
-	gconf, err := grantedConfig.Load()
-	if err != nil {
-		return err
-	}
-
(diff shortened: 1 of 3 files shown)


chrnorm (commented):
Looking good @Eddie023, could you please also add a test case to verify that works for a config like this:
```
[sso-session test]
sso_region    = us-west-2
sso_start_url = ...

[profile testing]
sso_session    = test
sso_account_id = ...
```

jordiup (approved):

jordiup (approved):

JoshuaWilkes:
Looks good @Eddie023 , seems like we should update the docs with an example of how to use sso-session.

If a user sets up multiple profile registry files best practice will be to add sso-session sections into a base config, so that it is not duplicated.

We don't have any logic here to deduplicate, I don't think deduplication would be necessary either because the last entry in the config file will be the one that is used

JoshuaWilkes (approved):

PR #359, opened 2023-02-08
AWS profile generation improvements

### LINKED ISSUE #270: `granted sso populate` should use the credential_process integration
Our credential_process integration is the smoothest UX for working with AWS SSO, as it allows users to use the `--profile` without being prompted for a second login. We should add support to `granted sso populate` and `granted sso generate` for it.

### LINKED ISSUE #340: [FR] add --sync option to sso populate that when used with --prefix would prune stale entries
We use granted to only populate our AWS config files with profiles that are currently available to the user. When new profiles are added, a `granted sso populate --prefix production-` will add those entries. My request is to add a `--sync` flag that would remove any old profiles that are no longer available that matched that profile.

`--sync` would:
* only be relevant if used in conjunction with the `--prefix` option
* be opt-in obviously, not changing any current behavior
* only remove stale entries where the prefix matches the one specified by `--prefix`

I am open to creating PR to add this functionality if the functionality would be accepted. The option could also be named `--prune`. Or it could be implemented as a new command to `sso` for `aws sso prune --prefix "foo"`. I would prefer it not be a separate command though as scanning our SSO profiles via the API already takes over 60 seconds, and doing a prune then a populate would take twice as long.

chrnorm:
Improvements to the `granted sso generate` and `granted sso populate` commands.

- use awsconfigfile package to generate AWS SSO profiles
- use Granted credential source by default
- shift CLI log messages relating to profile fetching to listSSOProfiles
- introduce `--sso-region` flag
- warn users of future breaking change if `--region` flag is used
- warn users if neither `--sso-region` nor `--region` is provided
- adds Common Fate as a supported source to generate profiles from. This allows users to generate profiles for privileged access roles from AWS SSO that they don't have persistent access to.
- adds a `--source` flag to indicate where to source profiles from. Currently the supported sources are `aws-sso` and `commonfate` - by default, `aws-sso` is used.
- adds a `--prune` flag which removes stale profiles from the local config file if they no longer exist in the source

Note that this changes the generated SSO profiles to use our credential_process integration by default. Users can opt-out of this by providing the `--no-credential-process` flag. Given that our credential process integration has been working well for several releases I don't consider this a breaking change.

Closes #270 
Closes #340 

These are the warnings printed if `--sso-region` isn't provided:

```
❯ granted sso generate https://example.awsapps.com/start
[!] Please specify the --sso-region flag: 'granted sso populate --sso-region us-east-1 https://example.awsapps.com/start'
[!] Currently, Granted defaults to using us-east-1 if this is not provided. In a future version, the --sso-region flag will be required to indicate the 'sso_region' field in generated profiles (https://github.com/common-fate/granted/issues/360)
```

```
granted sso generate --region ap-southeast-2 https://example.awsapps.com/start
[!] Please use --sso-region rather than --region.
[!] In a future version of Granted, the --region flag will be used to specify the 'region' field in generated profiles, rather than the 'sso_region' field. (https://github.com/common-fate/granted/issues/360)
```

To fix the warnings, use `--sso-region` rather than `--region`.


chrnorm:
This also closes #340

PR #374, opened 2023-02-28
Fix credential process for headless clients

meyerjrr:
Add bool config opt to specify if credential process is context. Returns error when no sso token is found in the cache when attempting to get creds

### DIFF
--- a/pkg/granted/sso.go
+++ b/pkg/granted/sso.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/AlecAivazis/survey/v2"
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/sso"
@@ -17,6 +18,7 @@ import (
 	"github.com/common-fate/clio/clierr"
 	"github.com/common-fate/granted/pkg/cfaws"
 	"github.com/common-fate/granted/pkg/securestorage"
+	"github.com/common-fate/granted/pkg/testable"
 	"github.com/urfave/cli/v2"
 	"gopkg.in/ini.v1"
 )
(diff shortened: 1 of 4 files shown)


meyerjrr:
Adds `granted sso login` to the granted binary, which will run an SSO cred flow to get a sso session token. Then saves to the keychain.

pkg/granted/sso.go
```diff
@@ -199,6 +201,57 @@ var PopulateCommand = cli.Command{
...
+	Flags: []cli.Flag{
+		&cli.StringFlag{Name: "sso-region", Usage: "Specify the SSO region"},
+		&cli.StringFlag{Name: "sso-start-url", Usage: "Specify the SSO start url"},
+	},
+	Action: func(c *cli.Context) error {
+		ctx := c.Context
+		ssoStartUrl := c.String("sso-start-url")
+
+		if ssoStartUrl == "" {
+			in1 := survey.Input{Message: "SSO Start Url"}
```
chrnorm, line 216:
```suggestion
			in1 := survey.Input{Message: "SSO Start URL"}
```

pkg/granted/sso.go
```diff
@@ -199,6 +201,57 @@ var PopulateCommand = cli.Command{
...
+
+		if cachedToken == nil {
+			newSSOToken, err := cfaws.SSODeviceCodeFlowFromStartUrl(ctx, *cfg, ssoStartUrl)
+			if err != nil {
+				return err
+			}
+
+			secureSSOTokenStorage.StoreSSOToken(ssoStartUrl, *newSSOToken)
+		} else {
+			clio.Info("Cached token found. Aborting SSO authentication flow.")
```
chrnorm, line 246: If a user wants to refresh their token because they are having login issues this might cause an undesired effect - I think will be easier just to remove checking the cache and always save a new token here.

chrnorm:
Thanks @Jasper-M - @sosheskaz has contributed an improvement in #418 to keep the old behaviour.

PR #376, opened 2023-03-01
CF-716: support for ~/.aws/sso/cache

jordiup:
## Describe your changes

**Problem**

Granted doesn’t yet have coverage of `sso/cache` 

```diff
~/.aws/sso/cache
+└── a092ca4eExample27b5add8ec31d9b.json
```

When running `aws sso login` credentials are stored in plain text here

This means that when running subsequent queries to `assume` the default behaviour will require you to re-authenticate (poor UX), rather than read from here

****Important distinction:**** `~/.aws/sso/cache` is different from `~/.aws/config` and `~/.aws/credentials`

Therefore we must add suport reading creds from `~/.aws/sso/cache`

****************Implementation Problem**************** 

This is fine for just one aws profile, but this may complicate things if a (power) user is logged in to more than one SSO session at once. IMO, this can be added in a subsequent PR. We’ll have to do a multi-lookup and specify an id

```diff
~/.aws/sso/cache
└── a092ca4eExample27b5add8ec31d9b.json
+└── anotherRandomStringSetOfCreds.json
+└── anotherRandomStringSetOfCreds.json
+└── anotherRandomStringSetOfCreds.json
```

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Issue and Documentation

https://linear.app/common-fate/issue/CF-716/granted-sso-generatepopulate-commands-should-fall-back-to-plaintext

## Testing

### remove tokens
granted sso-tokens clear

### now log in 
aws sso login --profile AWSAdministratorAccess-902101723122

### this adds an additional entry to ~/.aws/sso/cache
```diff
~/.aws/sso/cache
+└── a092ca4eExample27b5add8ec31d9b.json
```

### now try gen with the roles
```
go run ./cmd/granted/main.go sso generate --sso-region ap-southeast-2 [url]
```

### DIFF
--- a/pkg/granted/sso.go
+++ b/pkg/granted/sso.go
@@ -253,7 +253,7 @@ func (s AWSSSOSource) GetProfiles(ctx context.Context) ([]awsconfigfile.SSOProfi
 	if ssoToken == nil {
 		// if there is no valid token in the cache, check in ~/.aws/sso/cahe
 		if cfaws.SsoCredsAreInConfigCache() {
-			creds, err := cfaws.ReadCreds()
+			creds, err := cfaws.ReadPlaintextSsoCreds(s.StartURL)
 			if err != nil {
 				return nil, err
 			}
--- a/pkg/granted/sso.go
+++ b/pkg/granted/sso.go
@@ -277,7 +277,7 @@ func (s AWSSSOSource) GetProfiles(ctx context.Context) ([]awsconfigfile.SSOProfi
 
 	ssoClient := sso.NewFromConfig(*cfg)
 
-	// if the token is nill fetch it from config instead
+	// if the token is nil fetch it from config instead
 	var ssoProfiles []awsconfigfile.SSOProfile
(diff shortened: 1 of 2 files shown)


pkg/granted/sso.go
```diff
@@ -245,24 +247,43 @@ func (s AWSSSOSource) GetProfiles(ctx context.Context) ([]awsconfigfile.SSOProfi
...
+	} else {
+		accessToken = ssoTokenFromPlainText.AccessToken
 	}
-	secureSSOTokenStorage.StoreSSOToken(s.StartURL, *ssoToken)
 
 	clio.Info("listing available profiles from AWS IAM Identity Center...")
 
 	ssoClient := sso.NewFromConfig(*cfg)
 
+	// if the token is nill fetch it from config instead
```
chrnorm, line 280:
```suggestion
	// if the token is nil fetch it from config instead
```

chrnorm (approved):

PR #389, opened 2023-03-22
Add fallback support for any POSIX-compatible shell

jhuntwork:
As a last resort before erroring, try to determine a 'config' or 'profile' file to use via a POSIX-specific test.

[POSIX states regarding the ENV variable](https://pubs.opengroup.org/onlinepubs/9699919799/):

> This variable, when and only when an interactive shell is invoked, shall be
> subjected to parameter expansion (see Parameter Expansion) by the shell, and
> the resulting value shall be used as a pathname of a file containing shell
> commands to execute in the current environment. The file need not be
> executable.

## Motivation

My main shell is typically [mksh](http://www.mirbsd.org/mksh.htm). Even though it lacks many of the bells and whistles of bash and zsh, it is an adequate and comfortable interactive shell. By making all the common pieces POSIX-compliant, all the currently supported shells should continue to work and any additional shell that is POSIX-compliant should work as well.

This also requires some changes in the assume shim script. Even though some parts are zsh-specific and warnings can be ignored, the whole thing is sourced and so needs to be syntactically correct:

- Replace '<<<' with a simple heredoc.
- Replace the 'source' command with '.'
- Use parameter substitution for testing the existence of substrings
- Remove some other unnecessary bashisms
- Minor tweak, but replace uses of `echo` with `printf` since that tends to be safer and more explicit in formatting. A good summary of the difference is [here](https://www.etalabs.net/sh_tricks.html)

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Testing

1. Just testing the granted workflow works with these changes in your environment. I didn't add any additional tests, to the code base but am happy to do so with some guidance. Also happy to add any docs/comments as needed.


chrnorm (approved):
Thanks so much for the contribution @jhuntwork! 🙌 

I tested this in `fish`, `zsh` and `bash` and it works well for me. I ran into an error in `zsh`:
```
/usr/local/bin/dassume: line 197: return: can only `return' from a function or sourced script
```

so I've changed the `return` statement to be `exit`. These have the same effect AFAIK as we're trying to return a status code and exit the script.

PR #391, opened 2023-03-23
CLI based access request workflow

chrnorm:
### What changed?
Adds 
```
granted exp request latest
```

and 

```
granted exp request aws
```
commands for making Access Requests to AWS roles.

https://www.loom.com/share/f49279efc7464f268b49a0d7a1acbe94

### Why?
Improves the user experience when Granted is used with a Common Fate deployment. No impact to UX if Granted is used without CF - a user won't use these subcommands.

### How did you test it?
```bash
sudo make cli

dassume
# select a role which requires an Access Request to be made

dgranted exp request latest
# this should prompt you for a reason to access the same role
```

```bash
dgranted exp request aws
# should be able to select an account ID and role and then an access request is created
```

### Potential risks
Shouldn't affect any existing functionality as these are new subcommands

No - new feature

### Link to relevant docs PRs
https://github.com/common-fate/docs/pull/223


pkg/cache/migrations/20230316153630_init.down.sql
JoshuaWilkes, line 1: @chrnorm did you remove this dependency?
Seems like the cache migrations are unused

pkg/granted/exp/request/request.go
```diff
@@ -0,0 +1,580 @@
...
+		&awsCommand,
+		&latestCommand,
+	},
+}
+
+var awsCommand = cli.Command{
+	Name:  "aws",
+	Usage: "Request access to an AWS role",
+	Flags: []cli.Flag{
+		&cli.StringFlag{Name: "account", Usage: "the AWS account ID"},
```
JoshuaWilkes, line 46:
```suggestion
		&cli.StringFlag{Name: "account", Usage: "The AWS account ID"},
```

pkg/granted/exp/request/request.go
```diff
@@ -0,0 +1,580 @@
...
+		&cli.StringFlag{Name: "role", Usage: "the AWS role"},
```
JoshuaWilkes, line 47:
```suggestion
		&cli.StringFlag{Name: "role", Usage: "The AWS role"},
```

pkg/granted/exp/request/request.go
```diff
@@ -0,0 +1,580 @@
...
+		&cli.StringFlag{Name: "reason", Usage: "a reason for access"},
```
JoshuaWilkes, line 48:
```suggestion
		&cli.StringFlag{Name: "reason", Usage: "A reason for access"},
```

pkg/granted/exp/request/request.go
```diff
@@ -0,0 +1,580 @@
...
+			reason:  c.String("reason"),
+		})
+	},
+}
+
+var latestCommand = cli.Command{
+	Name:  "latest",
+	Usage: "Request access to the latest AWS role you attempted to use",
+	Flags: []cli.Flag{
+		&cli.StringFlag{Name: "reason", Usage: "a reason for access"},
```
JoshuaWilkes, line 63:
```suggestion
		&cli.StringFlag{Name: "reason", Usage: "A reason for access"},
```

JoshuaWilkes (changes requested):

JoshuaWilkes (changes requested):
I only had a comment about the sql migration files.

The rest looks fine to me

JoshuaWilkes:
Testing this, I was able to get through requesting access ok by following the prompts.

I had to first request access and have it be denied after the web based aws flow. (I expected this, but it's something we can look at improving/ autoprompting to request)

After requesting I was shown a prompt 

image
That said I had access, however because this command runs as a subcommand of granted, it is unable to write credentials to my terminal environment.

I know I needed to then run `assume` but a regular user may get lost at this point.
We can add another info login here with `run assume latest`

pkg/cache/migrations/20230316153630_init.down.sql
chrnorm: Thanks, I've cleaned this up

JoshuaWilkes (approved):

PR #394, opened 2023-03-29
Add disableusagetips setting and update mechanism

JoshuaWilkes:
## Describe your changes

- In response to user feedback, experienced users would like to be able to disable the usage tips presented when using assume.
- As a catch all, I have implemented a config flag which can be set to disable these tips from showing
- I have only disabled the tip that shows when you don't supply a profile name during assume, more can be added as requested.
- In order to make this and other simple config easy to update, I have added a generic `granted settings set` command which can set the value for string, int and bool config fields
- This way we can share with users a one liner to disable the tips `granted settings set -s DisableUsageTips -vb` will configure their installation to not show the tip.

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## Issue and Documentation

-

## Testing

1. `make cli`
1. try running `dgranted settings set -s DisableUsageTips -vb` followed by `assume` do not supply a profile name, expect not to see a usage tip.
2. try running `dgranted settings set -s DisableUsageTips -vb false`  followed by `assume` do not supply a profile name, expect to see the prompt with the tip on usage.

3. run  `dgranted settings set` and see that you can update other settings

- [ ] I have performed a self-review of my code.
- [ ] I have commented my code, particularly in hard-to-understand areas.

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -301,7 +301,7 @@ func AssumeCommand(c *cli.Context) error {
 	getConsoleURL := !assumeFlags.Bool("env") && (assumeFlags.Bool("console") || assumeFlags.Bool("active-role") || assumeFlags.String("service") != "" || assumeFlags.Bool("url"))
 
 	// this makes it easy for users to copy the actual command and avoid needing to lookup profiles
-	if showRerunCommand {
+	if !cfg.DisableUsageTips && showRerunCommand {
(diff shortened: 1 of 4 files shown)


chrnorm:
Because we're parsing a string from the CLI, I have simplified the `--value-bool`, `--value-str` and `--value-int` into a single flag `--value`. This will help avoid a problem where a user accidentally gives the wrong flag for a particular config value

chrnorm (approved):

PR #400, opened 2023-04-11
Reintroduce fallback support for any POSIX-compatible shell

chrnorm:
This was originally merged to `main` in @jhuntwork's PR #389, however for some reason it caused some of our GitHub Actions tests to fail. I've reverted the PR temporarily and am introducing it again on this branch to evaluate why the tests were failing.

------------

* Add fallback support for any POSIX-compatible shell

As a last resort before erroring, try to determine a 'config' or 'profile' file to use via a POSIX-specific test.

[POSIX states regarding the ENV variable](https://pubs.opengroup.org/onlinepubs/9699919799/):

> This variable, when and only when an interactive shell is invoked, shall be
> subjected to parameter expansion (see Parameter Expansion) by the shell, and
> the resulting value shall be used as a pathname of a file containing shell
> commands to execute in the current environment. The file need not be
> executable.

## Motivation

My main shell is typically [mksh](http://www.mirbsd.org/mksh.htm). Even though it lacks many of the bells and whistles of bash and zsh, it is an adequate and comfortable interactive shell. By making all the common pieces POSIX-compliant, all the currently supported shells should continue to work and any additional shell that is POSIX-compliant should work as well.

This also requires some changes in the assume shim script. Even though some parts are zsh-specific and warnings can be ignored, the whole thing is sourced and so needs to be syntactically correct:

- Replace '<<<' with a simple heredoc.
- Replace the 'source' command with '.'
- Use parameter substitution for testing the existence of substrings
- Remove some other unnecessary bashisms
- Minor tweak, but replace uses of `echo` with `printf` since that tends to be safer and more explicit in formatting. A good summary of the difference is [here](https://www.etalabs.net/sh_tricks.html)

* change 'return' to 'exit'

### DIFF
--- a/pkg/alias/alias.go
+++ b/pkg/alias/alias.go
@@ -30,9 +30,9 @@ func init() {
 }
 
 const fishAlias = `alias assume="source /usr/local/bin/assume.fish"`
-const defaultAlias = `alias assume="source assume"`
+const defaultAlias = `alias assume=". assume"`
 const devFishAlias = `alias dassume="source /usr/local/bin/dassume.fish"`
-const devDefaultAlias = `alias dassume="source dassume"`
+const devDefaultAlias = `alias dassume=". dassume"`
 
 func GetDefaultAlias() string {
 	if build.IsDev() {
@@ -77,7 +77,9 @@ func GetShellFromShellEnv(shellEnv string) (string, error) {
 
 	} else if strings.Contains(shellEnv, "zsh") {
 		return "zsh", nil
-
+	} else if strings.Contains(shellEnv, "sh") {
+		// Fallback to POSIX
+		return "posix", nil
 	} else {
 		return "", fmt.Errorf("we couldn't detect your shell type (%s). Please follow the steps at https://docs.commonfate.io/granted/internals/shell-alias to assume roles with Granted", shellEnv)
 	}
@@ -96,6 +98,8 @@ func GetShellAlias(shell string) (Config, error) {
 		file, err = shells.GetBashConfigFile()
 	case "zsh":
 		file, err = shells.GetZshConfigFile()
+	case "posix":
+		file, err = shells.GetPosixConfigFile()
 	default:
 		err = &ErrShellNotSupported{Shell: shell}
 	}
--- a/pkg/shells/find_config.go
+++ b/pkg/shells/find_config.go
@@ -1,9 +1,35 @@
 package shells
 
 import (
+	"fmt"
 	"os"
 	"path"
+	"path/filepath"
+	"strings"
 )
+
+func GetPosixConfigFile() (string, error) {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return "", err
+	}
+	file := os.Getenv("ENV")
+	if file == "" {
+		return "", fmt.Errorf("cannot determine a shell config file. Please specify one using the ENV variable")
+	}
+	if strings.HasPrefix(file, "~") {
+		file = filepath.Join(home, file[1:])
+	}
+	// check that the file exists; create it if not
+	if _, err := os.Stat(file); os.IsNotExist(err) {
+		f, err := os.Create(file)
+		if err != nil {
+			return "", err
+		}
+		defer f.Close()
+	}
+	return file, nil
+}
 
 func GetFishConfigFile() (string, error) {
 	home, err := os.UserHomeDir()
--- a/scripts/assume
+++ b/scripts/assume
@@ -1,10 +1,12 @@
-#!/bin/bash
+#!/bin/sh
 
 #GRANTED_FLAG - what assumego told the shell to do
 #GRANTED_n - the data from assumego
 
 # pass an environment variable to the Go binary if the Granted alias hasn't been configured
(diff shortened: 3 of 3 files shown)


jhuntwork:
> Any chance this is because of the 'exit' instead of the 'return' ? As I understand it, the script is being sourced, which means if you call 'exit', you are actually exiting the parent shell that sources the script.
> 
> If you have any guidance on replicating these tests locally, I'd be happy to dig into the failure a bit too.



> Any chance this is because of the 'exit' instead of the 'return' ? As I understand it, the script is being sourced, which means if you call 'exit', you are actually exiting the parent shell that sources the script.
> 
> If you have any guidance on replicating these tests locally, I'd be happy to dig into the failure a bit too.

Hmm, nope, nevermind, that would apply to functions. In the main body of a script that you source it would be effectively the same.

chrnorm:
Here's a print of the relevant logs from GitHub Actions:

```
Run SHELL=/bin/bash dassume -auto-configure-shell
  SHELL=/bin/bash dassume -auto-configure-shell
  shell: /usr/bin/bash -e {0}
  env:
    TEST_ACCESS_KEY_ID: 1[2](https://github.com/common-fate/granted/actions/runs/4670007054/jobs/8269271951?pr=400#step:8:2)[3](https://github.com/common-fate/granted/actions/runs/4670007054/jobs/8269271951?pr=400#step:8:3)[4](https://github.com/common-fate/granted/actions/runs/4670007054/jobs/8269271951?pr=400#step:8:4)[5](https://github.com/common-fate/granted/actions/runs/4670007054/jobs/8269271951?pr=400#step:8:5)78910abcdefghij
    TEST_ACCESS_KEY: 1234578910abcdefghijklmnopqrstuvwxyz1234
    TEST_REGION: ap-southeast-2

25l? Please select the profile you would like to assume:  [Use arrows to move, type to filter]
> test
25h[✘] EOF
Error: Process completed with exit code 1.
```

Usually the command `assume -auto-configure-shell` does not cause the interactive prompt to open. I am wondering if the changes in this PR caused the arguments to not be parsed properly anymore, so that the Go program is just receiving `assume` rather than `assume -auto-configure-shell`?

jhuntwork:
So I believe this is to do with my misreading of the initial code and not understanding the goal of the alias. The challenge is that POSIX doesn't define a way to determine if a script is being sourced or executed directly. `bash` provides the extra `BASH_SOURCE` variable as was used here before. And other shells have other mechanisms. It seems the easiest quasi-portable method is to account for how `$0` is treated in different shells when the script is sourced. 

Here's how that breaks down for 4 different shells:

Save this scriptlet as `/usr/local/bin/dassume`:

```sh
#!/bin/sh
# Check the type of the basename of "$0"
printf 'type of $0: %s\n' "$(type -- "${0##*/}")"
printf '$0: %s\n' "$0"
```

Then without the alias set up, notice the output in each of the following

**bash**
```
$ dassume
type of $0: dassume is /usr/local/bin/dassume
$0: /usr/local/bin/dassume
```

**zsh**
```
% dassume
type of $0: dassume is /usr/local/bin/dassume
$0: /usr/local/bin/dassume
```

**mksh**
```
$ dassume
type of $0: dassume is /usr/local/bin/dassume
$0: /usr/local/bin/dassume
```

**dash**
```
$ dassume
type of $0: dassume is /usr/local/bin/dassume
$0: /usr/local/bin/dassume
```

So far so good, all treat the direct script the same. Now let's set up the alias on each, running `alias dassume='. /usr/local/bin/dassume'` and then run the alias.

**bash**
```
$ dassume
-bash: type: -bash: not found
type of $0:
$0: -bash
```

**zsh**
```
% dassume
type of $0: dassume is an alias for . /usr/local/bin/dassume
$0: /usr/local/bin/dassume
```

**mksh**
```
$ dassume
type of $0: -mksh not found
$0: -mksh
```

**dash**
```
$ dassume
type of $0: --: not found
-dash: not found
$0: -dash
```

They all handle it slightly differently, but one thing to note is that `$0` is usually the name of the login shell and the type command fails. Except for `zsh`. But in that case, the type command shows that the command is definitely an alias. I think the way to go is to either test for failure of the type command, or that the 'alias' keyword is present.

The following code catches the case correctly for all 4 shells:
```sh
#!/bin/sh
_this_type=$(type -- "${0##*/}" 2>&1)

# If the type test failed, this is probably a sourced
# file because $0 will be the login shell, e.g., -bash,
# and the output will typically contain 'not found'
# In the case of zsh, the output will contain the word 'alias'.

if [ "${_this_type#*not found}" != "$_this_type" ] ||
        [ "${_this_type#*alias}" != "$_this_type" ]; then
    printf 'This file was sourced\n'
else
    printf 'This file was executed directly\n'
fi
```

jhuntwork:
I think https://github.com/common-fate/granted/pull/438 fixes the issues with this code. I see that this also needs to be rebased and have its conflicts fixed. Would you like me to do that or how do you want to proceed?

jhuntwork:
I ended up rebasing #438. I think that has everything needed there, and if we accept that one, this one can be closed.

Issue #403, opened 2023-04-13
Reduce/Eliminate User Prompts for Keychain Access


lyoung-confluent:
When using `granted`/`assume` users may experience prompts for their password to access the Keychain items, particularly on macOS:
<img width="546" alt="image" src="https://user-images.githubusercontent.com/91491244/231655585-0a9b82ed-b423-4f86-abd6-76e1e84a0ad1.png">
This can be reproduced easily on macOS by running `assume` for a valid AWS SSO profile then running `granted sso-tokens clear --all` to clear out the entry.

I suspect this is related to some combination in an oddity with the code signing of the 2 separate binaries, the ACLs on the keychain items or the significantly older version of https://github.com/99designs/go-keychain used by `granted`, however more investigation is required.

lyoung-confluent:
Upon further investigation using https://github.com/its-a-feature/LockSmith (see also [Lock Picking the macOS Keychain talk at OBTS 5.0](https://www.youtube.com/watch?v=jKE1ZW33JpY)), we can see the ACLs on a keychain item created using the signed `v0.10.0` release:
```
Account:       https://<snip>.awsapps.com/start
Label:         granted-aws-sso-tokens
Service:       granted-aws-sso-tokens
Creation Date: 2023-04-12 21:08:51
Modify Date:   2023-04-12 21:08:51
Class:         genp
Owner Authorizations based on ACLAuthorizationPartitionID
Owner Authorizations:
    ACLAuthorizationEncrypt
    ACLAuthorizationDecrypt
    ACLAuthorizationDerive
    ACLAuthorizationExportClear
    ACLAuthorizationExportWrapped
    ACLAuthorizationMAC
    ACLAuthorizationSign
    ACLAuthorizationIntegrity
    ACLAuthorizationPartitionID
------------ACLS------------
    --- Entry 0
        Description: (null)
        All applications are trusted
        Authorizations:
             ACLAuthorizationEncrypt
    --- Entry 1
        Description:
        Trusted App: /usr/local/Cellar/granted/0.10.0/bin/assumego
            Requirement String: identifier assumego and anchor apple generic and certificate 1[field.1.2.840.113635.100.6.2.6] /* exists */ and certificate leaf[field.1.2.840.113635.100.6.1.13] /* exists */ and certificate leaf[subject.OU] = VXC4RPMSUJ
        Authorizations:
             ACLAuthorizationDecrypt
             ACLAuthorizationDerive
             ACLAuthorizationExportClear
[...]

PR #404, opened 2023-04-13
Replace assumego with a symlink to granted fixing keychain ACL issues

lyoung-confluent:
## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## Issue and Documentation
This change merges the `assumego` and `granted` implementation into a single binary with the behavior swapping from one to the other based the name of the executable used (i.e. `os.Args[0]`). The `assumego` binary is replaced with a symlink to the `granted` binary instead. This allows a single binary (`granted`) to be accessing the keychain items eliminating the prompts for access that users experience (see also issue #403 for more details/investigation). 

**This will require additional changes to the production build/sign/release pipeline (not part of OSS repo), and to the published installers like the [homebrew formula](https://github.com/common-fate/homebrew-granted/blob/main/Formula/granted.rb) (ex: `bin.install` becomes `bin.install_symlink`).**

## Testing
On macOS use `dassume` for a valid AWS SSO profile then run `dgranted sso-tokens clear --all` to clear out the entry and observe that no keychain prompt occurs.

- [x] New and existing unit tests pass with my changes.


chrnorm (approved):
I'm going to merge and include this in our next release - thanks so much for your work looking into this @lyoung-confluent. I need to research how we can bundle symlinks for Linux and Windows into the release zip folder. In the meantime, I will update the brew formula to use the symlink and simply build the binary twice for Windows and Linux. We should certainly look at optimising this but I am keen to get this improvement in ASAP as it improves the MacOS user experience a fair deal.

PR #409, opened 2023-04-20
Append a single array element for with field

JoshuaWilkes:
### What changed?
Corrected the data type for with field when making a request for a rule which had multiple options for both permissionset and account

### Link to relevant docs PRs


chrnorm (approved):
Tested and working locally for me

PR #418, opened 2023-05-29
Add headless mode cfg flag to control auto-login

sosheskaz:
### What changed?

Instead of prompting the user to log in, log in automatically, unless they specifically opt-out by configuring `HeadlessMode` in the granted config.

### Why?

This was a regression in 0.9.0 for users using `credential-process` on desktop environments, due to removing this behavior for headless users.

### How did you test it?

1. Link up dgranted and dassume, reconfigure my AWS profiles to use those binaries.
2. Purge granted tokens from keychain.
3. Run `aws --profile <profile> iam list-account-aliases`.
  1. It automatically opens my configured browser and logs me in. This confirms that the path for "non-headless" mode works as expected.
4. Purge granted tokens from keychain again.
5. Run `dgranted settings set -s HeadlessMode -v true`
6. Run `aws --profile <profile> iam list-account-aliases`.
  7. It prompts me to run the `granted sso login` command

### Potential risks

This may be initially breaking for headless users until they update configuration

No; it should be a minor release because it may be breaking for headless users.

### Link to relevant docs PRs


sosheskaz:
My impression of granted, as a tool, is that it is predominantly for humans, which typically use desktop environments. Is that not the case?

From a UX perspective, it is probably better to target the wider audience's use-case. Especially given, IME, headless users/envs will tend to be
1. more familiar with the underlying CLI tooling
2. controlled by automated config mgmt systems of some sort
and thus it would easier to roll out the configuration change, as compared to humans, which are... humans. Most users are not advanced, and do not tinker with default settings, and so IMO it is valuable that the default settings be well-suited to those users.

----

That said, the regression on headless machines is far worse, and the message as Chris outlined would probably be sufficient for my purposes.

----

Something that might make life a bit easier for me, would be if this could be overridden via CLI flag (i.e. `granted credential-process --auto-login`. That way, I could probably roll it out automatically via registry.

lorengordon:
> Something that might make life a bit easier for me, would be if this could be overridden via CLI flag (i.e. granted credential-process --auto-login. That way, I could probably roll it out automatically via registry.

A CLI flag makes sense to me also. That would allow for the case where maybe you need to support both headless and desktop modes. The Granted config could specify or default to `AutoLogin: false`, which would work fine for a headless profile. And I could have a second aws-cli profile that uses `--auto-login` for when I am interactive.

chrnorm:
> My impression of granted, as a tool, is that it is predominantly for humans, which typically use desktop environments. Is that not the case?

Indeed - I certainly agree. To clarify here, I use "headless" to refer to the use-case where a human user is using an interactive terminal which does not permit a web browser to be opened automatically, such as on an EC2 instance.

`--auto-login` sounds like an excellent idea. @sosheskaz let us know if you have time to update this PR, otherwise one of the core maintainers can add this on top of your existing work here. Thankyou!

sosheskaz:
I changed it to use a flag exclusively. If it is defined both in the config, and in the CLI, then that creates some ambiguity around behavior that I think would be better avoided.

It seems to me like profiles + CLI flags ought to be a good way to resolve this universally.

It does mean that headless and non-headless users effectively cannot share registries for their login users (not for every profile), but given the variance that registries offers, that problem can be solved without too much legwork using this implementation.

Test method was mostly the same, but I used these two commands:

```bash
gsed 's/ granted/ dgranted/g' -i ~/.aws/config
gsed 's/ credential-process/ credential-process --auto-login/g' -i ~/.aws/config
```

lorengordon:
Saw an email this was merged and went to test it. Looks like it's not in a release yet? Slightly disconnected lately, realizing now the merge was just a couple days ago, so no rush. Just that I got prompted with the login error, and wanted to update to take advantage of this option! :D

PR #429, opened 2023-06-06
Add support for MFA with IAM credentials

mfzl:
### What changed?
A new assumer is added with support for MFA

### Why?
MFA does not work with IAM credentials

### DIFF
--- a/pkg/cfaws/assumer_aws_iam_mfa.go
+++ b/pkg/cfaws/assumer_aws_iam_mfa.go
@@ -1,0 +1,89 @@
+package cfaws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/common-fate/clio"
+	"github.com/common-fate/granted/pkg/securestorage"
+	"gopkg.in/ini.v1"
+)
+
+type AwsIamMfaAssumer struct {
+	AwsIamAssumer
+}
+
+func (aia *AwsIamMfaAssumer) AssumeTerminal(ctx context.Context, c *Profile, configOpts ConfigOpts) (aws.Credentials, error) {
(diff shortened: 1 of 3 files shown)


JoshuaWilkes:
Hey @mfzl Thanks for this PR, i'm wondering why the original AWS IAM assumer did not work for you, it does have some logic to prompt for MFA.

I see there is a difference between how it operates when the profile has a parent.

In your situation, are you using a base profile and a child profile or using a profile which has the MFASerial config on it?

Maybe we can fix the issue causing the original IAM assumer not to work.

I'm inclined to think that this also needs to set the aro.SerialNumber when there is no parent profile.

```
			// set the token provider up
			aro.TokenProvider = MfaTokenProvider
			aro.Duration = configOpts.Duration

			// If the mfa_serial is defined on the root profile, we need to set it in this config so that the aws SDK knows to prompt for MFA token
			if len(c.Parents) > 0 {
				if c.Parents[0].AWSConfig.MFASerial != "" {
					aro.SerialNumber = aws.String(c.Parents[0].AWSConfig.MFASerial)

				}
			}
```

mfzl:
Hey @JoshuaWilkes, thanks for looking into this.

This can be handled in the AWS IAM assumer as well. My profile is basically
```
[profile xyz]
credential_process    = granted credential-process --profile xyz
mfa_serial            = arn:aws:iam::xxx
region                = us-east-1
```
You guessed right about the current assumer, it's making the assumption that user assumes a role in the child profile so it's not triggered when using the credentials directly with MFA using the root profile (no parent)

This can be fixed in the current AWS IAM assumer as well. I felt it's safer to leave the existing code as is without introducing token caching and behavior change but it can be made to work. 

In the case of updating the current assumer this logic will go into 

```go
	if c.HasSecureStorageIAMCredentials {
		secureIAMCredentialStorage := securestorage.NewSecureIAMCredentialStorage()
		crendentials, err := secureIAMCredentialStorage.GetCredentials(c.Name)
		if err != nil {
		    return credentials, err
		}
                if c.AWSConfig.MFASerial == "" {
		    return credentials, nil
	        }

		// Handle MFA code
	}

```

chrnorm (approved):

PR #435, opened 2023-06-19
Make assume script a little more robust

alexjurkiewicz:
This script might be sourced from another one which uses `set -u`.
Ensure we don't unconditionally access optional environment variables.

(ignore whitespace to understand the diff better)


chrnorm (approved):
Thankyou @alexjurkiewicz! 🙌 I ran a quick test of this locally it's working well for me on ZSH

Issue #246, opened 2022-09-09
`assume` CLI doesn't work with fish on Apple Silicon MacOS


chrnorm:
Apple Silicon devices seem to default to installing things in `/opt/homebrew` which doesn't match the shell alias that we add.

Additionally, we shouldn't need the `#!/bin/fish` shebang because the script is being sourced.

admackin:
I believe `brew --prefix` is what is needed here, instead of hardcoding to `/usr/local/bin`

PR #437, opened 2023-06-22
use brew --prefix instead of hardcoded path for fish alias

admackin:
the hardcoded path doesn't work on brew installs on apple silicon (or if someone specifies an alternate location for brew)

### What changed?

Used `brew --prefix` to get root homebrew location

### Why?

as reported in #246 the alias installation doesn't work for `fish` on Apple Silicon (M1/M2) since the homebrew binaries are installed in `/opt/homebrew` rather than `/usr/local`

### How did you test it?

Manually inserted the same change in my local `config.fish` (on Apple silicon). 

### Potential risks

not e2e-tested on intel macs (but I have verified that `brew --prefix` command works identically on intel-based Brew, which I also have installed)

yes, it's a small bugfix that should make things work for all people who use Fish on Apple Silicon, and be unnoticed by everyone else.

### DIFF
--- a/pkg/alias/alias.go
+++ b/pkg/alias/alias.go
@@ -29,9 +29,9 @@ func init() {
 	_, color.NoColor = os.LookupEnv("NO_COLOR")
 }
 
-const fishAlias = `alias assume="source /usr/local/bin/assume.fish"`
+const fishAlias = `alias assume="source (brew --prefix)/assume.fish"`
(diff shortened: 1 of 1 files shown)


chrnorm (approved):
Thankyou for the contribution @admackin!

I'm conscious that this might cause problems for non-brew users on MacOS, but given that our install instructions only reference `brew` this should be a non-issue.

Issue #371, opened 2023-02-22
Improve MFA login experience by caching session tokens


smelchior:
It would be great if granted could cache the session tokens after an MFA login that are used to assume roles in a non-SSO setup.
This would be for an aws cli config like:
```
[xy-base]
aws_access_key_id     = XXX
aws_secret_access_key = YYYY

[xy-operations]
role_arn       = arn:aws:iam::XXX:role/XY
mfa_serial     = arn:aws:iam::XXXX:mfa/XY
source_profile = xy-base
```

In this case i am currently prompted for my MFA each time i run the assume call and i am also limited to the role session limit time (which is mostly 1h)
I was using https://github.com/trek10inc/awsume/ before, and they implemented something to cache the credentials returned by the get-session-token call for the MFA authentication (https://github.com/trek10inc/awsume/blob/master/docs/general/overview.md?plain=1#L56) This makes life a lot easier when you have to assume the role in multiple shell sessions or want to also open it in the browser.

This was partially discussed in this https://commonfatecommunity.slack.com/archives/C0356DGLXME/p1677055552697469 +this https://commonfatecommunity.slack.com/archives/C0356DGLXME/p1675985537292519 (with a slightly different use-case) Slack thread

orenmazor:
this would be really great! I jump between accounts all day and having to re-pull the MFA is kinda a bummer

PR #447, opened 2023-07-26
add a caching mechanism to assume

orenmazor:
### What changed?

Inspired by https://github.com/common-fate/granted/issues/371, I duplicated the credential caching mechanism from credential_process.go

### Why?

In my use case, I have the following setup, where I use 1password cli as the credentials source:

```
[profile oren]
region=us-east-1
credential_process=op --account agilebits.1password.com plugin run -- aws configure export-credentials

[profile someaccount]
source_profile=oren
role_arn=arn:aws:iam::x:role/sudo
mfa_serial=arn:aws:iam::x:mfa/oren.mazor
region=us-east-1
```

my issue is that I've got about 20 accounts and all of them require entering MFA. this is a pain in the butt and I'd like granted to cache the resulting MFA credentials

### How did you test it?

I've deployed this on my own machine and have been working with this today in both terminal and console features.

so far so good!

### Potential risks

I don't know enough about other folks use cases with granted. I'm not sure if I broke something here, so before I spend any more time DRYing things up (the same code is in credential process, assume terminal and assume console), I wanted to get some feedback!

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -28,6 +28,7 @@ import (
 	"github.com/common-fate/granted/pkg/console"
 	"github.com/common-fate/granted/pkg/forkprocess"
 	"github.com/common-fate/granted/pkg/launcher"
+	"github.com/common-fate/granted/pkg/securestorage"
 	"github.com/common-fate/granted/pkg/testable"
 	cfflags "github.com/common-fate/granted/pkg/urfav_overrides"
 	"github.com/fatih/color"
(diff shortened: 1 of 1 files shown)


Eddie023:
Hi @orenmazor Thanks for contributing to granted! 
Looks like you have added a top-layer caching mechanism to the assume command. However, there are multiple types of assumer some of which doesn't/cannot use `securestorage.NewSecureSessionCredentialStorage()`. I have added a fix in this [PR ](https://github.com/common-fate/granted/pull/474) where we instead cache instead the IAM_ASSUMER so that you do not have to reenter the MFA code if the current credential is valid.

Issue #458, opened 2023-08-11
granted sso generate/populate - TooManyRequestsException


rajkamalhm:
Generating sso config returns 429 status code from AWS - https://docs.aws.amazon.com/singlesignon/latest/PortalAPIReference/API_ListAccountRoles.html#API_ListAccountRoles_Errors

```
> granted sso populate --no-credential-process --profile-template "{{ .AccountName }}/{{ .AccountID }}" --sso-region eu-west-1 https://x-xxxxxxx.awsapps.com/start
[i] listing available profiles from AWS IAM Identity Center...  
[✘] operation error SSO: ListAccountRoles, exceeded maximum number of attempts, 3, https response error StatusCode: 429, RequestID: 85c43326-ebfb-4d68-b496-5558cbc9064c, TooManyRequestsException: HTTP 429 Unknown Code```

karlatkinson:
I'm seeing this on generate too. 
`granted sso generate --sso-region eu-west-1 https://xxxxxxxx.awsapps.com/start`
We have a large number of accounts and roles within them (around 200 accounts with around 5 roles in each)

PR #462, opened 2023-08-23
implements backoff in sso.go, avoids 429s when too many roles are present

c0nfleis:
### What changed?
Implemented backoffs when running listAccountRoles in sso.go

### Why?
429's when too many roles are present

### How did you test it?
I ran locally using make cli, and tested against real use case that was returning 429 when using main

### Potential risks
Possible still not enough attempts, but this number got me through

No

### DIFF
--- a/pkg/granted/sso.go
+++ b/pkg/granted/sso.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
+	"time"
 
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


egibs:
Addresses: https://github.com/common-fate/granted/issues/458

shwethaumashanker:
Thank you so much @c0nfleis for implementing the fix! We're really grateful for your work. Just a heads up, we've also just finished implementing a fix that rate limits the AWS SSO API call to [AWS's max throttle limit](https://docs.aws.amazon.com/singlesignon/latest/userguide/limits.html) and added a progress bar too #463

chrnorm:
Closing as we've just merged #463 - we opted for the implementation in that PR because it has a progress bar which should improve UX for large org sizes. This will be included in the next Granted release. 

As mentioned though we're hugely grateful for your contribution here @egibs - let us know how the next release works for you and contributions which further improve functionality here are very much welcomed.

Issue #263, opened 2022-09-26
The Credential Process assumer should only set AWS_PROFILE, not other variables


chrnorm:
If users have a [credential_process](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-sourcing-external.html) configured, the AWS CLI automatically handles refreshing credentials when they expire.

```
[profile my-profile]
credential_process = some-other-cli
```

However, when running `assume my-profile` the AWS CLI will not automatically refresh credentials. This is because we are exporting `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, and `AWS_SESSION_TOKEN`.

We should detect if a `credential_process` is in use and then only set `AWS_PROFILE`. We should *not* set any of the env vars listed above, as they cause the AWS CLI to not automatically refresh credentials.

tomharrisonjr:
@chrnorm -- this would be a really valuable feature, potentially solving several problems in one shot, I think:

* needless refresh, as noted in the description
* multiple terminal windows open with env vars that report `AWS_SESSION_EXPIRATION` relative to their last refresh, rather than the most recent one
* incompatibilities with tools that are configured to use profiles ... but pick up the env vars instead (and again, could be multiples if you have multiple terminal windows open

avadhanij6si:
+1 to this request. We have been trying to adopt this tool org wide, and having to `--export` on every terminal window is painful. It would be great \if the option could only populate the credentials file and set `AWS_PROFILE` variable.

gshpychka:
The docs [say](https://docs.commonfate.io/granted/recipes/credential-process#:~:text=By%20default%2C%20only%20the%20AWS_PROFILE%20environment%20variable%20is%20exported%20when%20you%20run%20assume%20%3Ccredential%2Dprocess%2Dprofile%3E) that this is already the case - should this be closed?

> By default, only the AWS_PROFILE environment variable is exported when you run assume <credential-process-profile>

CodyDunlap:
The docs may say it, but that's not what appears to be happening. I do not have either env variable set but when I run something using the AWS SDK after executing `assume my-profile` I get the following warning from AWS SDK:

```
@aws-sdk/credential-provider-node - defaultProvider::fromEnv WARNING:
    Multiple credential sources detected: 
    Both AWS_PROFILE and the pair AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY static credentials are set.
[...]

PR #467, opened 2023-09-01
Do not export aws credentials with credential_process_assumer

Eddie023:
### What changed?

Solves: https://github.com/common-fate/granted/issues/263

### Link to relevant docs PRs


pkg/assume/assume.go
```diff
@@ -451,16 +451,29 @@ func AssumeCommand(c *cli.Context) error {
...
 		}
+
 		// DO NOT REMOVE, this interacts with the shell script that wraps the assume command, the shell script is what configures your shell environment vars
 		// to export more environment variables, add then in the assume and assume.fish scripts then append them to this output preparation function
 		// the shell script treats "None" as an empty string and will not set a value for that positional output
+
+		// If the profile uses "credential_process" to source credential externally then do not set accessKeyId, secretAccessKey, sessionToken
+		// so that aws cli automatically refreshes credential when they expire.
+		if profile.ProfileType == "AWS_CREDENTIAL_PROCESS" {
+			output := PrepareStringsForShellScript([]string{"", "", "", "", region, sessionExpiration, profile.AWSConfig.SSOStartURL, profile.AWSConfig.SSORegion, profile.AWSConfig.SSORoleName, profile.AWSConfig.SSORegion, profile.AWSConfig.SSOAccountID})
```
JoshuaWilkes, line 462: In this code I don't think you should use returning the values for ssoStartURL etc as per the output when teh sso flag is used.

The reason we return these for the sso flag is because the SSO flag is used when you don;t have a profile configured already, and these flags help us to re assume the same config again later.

for profiles with a credential process, you should simply do the following

```suggestion
			output := PrepareStringsForShellScript([]string{"", "", "", profile.Name, region, "", "", "", "", "", ""})
```

JoshuaWilkes (changes requested):

JoshuaWilkes (dismissed):

JoshuaWilkes (commented):
There is potential for this to be an issue if you need to use assume with tooling that does not support credential process.
I also think this is a reasonably significant breaking change in behaviour.

My feeling is that it needs a workaround or more consultation in the community slack

JoshuaWilkes (approved):

chrnorm (dismissed):
Based on feedback from the community, let's add a flag `--export-all-env-vars` which will retain the current behaviour

pkg/assume/assume.go
```diff
@@ -451,16 +451,40 @@ func AssumeCommand(c *cli.Context) error {
 		if assumeFlags.String("exec") != "" {
 			return RunExecCommandWithCreds(assumeFlags.String("exec"), creds, region)
 		}
+
+		if assumeFlags.Bool("export-all-env-vars") {
```
JoshuaWilkes, line 455:
```suggestion
		if profile.RawConfig.HasKey("credential_process") assumeFlags.Bool("export-all-env-vars") {
```

JoshuaWilkes (changes requested):

pkg/assume/entrypoint.go
```diff
@@ -45,6 +45,7 @@ func GlobalFlags() []cli.Flag {
 		&cli.StringFlag{Name: "account-id", Usage: "Use this in conjunction with --sso, the account-id"},
 		&cli.StringFlag{Name: "role-name", Usage: "Use this in conjunction with --sso, the role-name"},
 		&cli.StringFlag{Name: "save-to", Usage: "Use this in conjunction with --sso, the profile name to save the role to in your AWS config file"},
+		&cli.BoolFlag{Name: "export-all-env-vars", Usage: "Exports all available credentials to the terminal"},
```
JoshuaWilkes, line 48:
```suggestion
		&cli.BoolFlag{Name: "export-all-env-vars", Usage: "Exports all available credentials to the terminal when used with a profile configured for credential-process. Without this flag, only the AWS_PROFILE will be configured"},
```

JoshuaWilkes: update comment and logic

JoshuaWilkes (approved):

dlambda:
Would be nice to have a short version `-x`(?) for `--export-all-env-vars` since exporting the creds for downstream processes (e.g. containers) is the way I've mainly been using `assume`.

PR #469, opened 2023-09-01
pass the output of getCallerIdentity as the name of the federated user

Eddie023:
### Link to relevant docs PRs

### DIFF
--- a/pkg/cfaws/assumer_aws_iam.go
+++ b/pkg/cfaws/assumer_aws_iam.go
@@ -125,7 +125,12 @@ func getFederationToken(ctx context.Context, c *Profile) (aws.Credentials, error
 
 	client := sts.NewFromConfig(cfg)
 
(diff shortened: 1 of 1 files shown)


JoshuaWilkes (approved):

matthewhembree:
I would have preferred that federation token name be composed of splitting `caller.Arn` instead of using `caller.UserId`. Something like a `strings.SplitAfter( caller.Arn, ":user/" )`

The ARN will have the canonical username. It's much clearer than seeing the abstraction of `AIDA...` as the username in CloudTrail.

But at least we now have a determinate name, so that my console preferences will persist. Thanks!!

izanagi1995:
@Eddie023 @JoshuaWilkes This is great, but it is almost unusable for Cloudtrail audits where we want to query by username. Now we instead have the User ID in the logs.

PR #512, opened 2023-10-12
Add Waterfox as a browser option

dowster:
### What changed?

Added options to configure Waterfox as a browser and prompt for install of the Firefox extension into Waterfox. 

### Why?

I had set up Waterfox originally by just providing the path to Waterfox instead of Firefox, but if a user had Firefox already installed to its default path, this might not have worked. 

### How did you test it?

Tested on MacOS
1. built the CLI locally with `PREFIX=~/.local make cli`
2. ran `dgranted browser set`
3. selected `Waterfox`
4. ran `dassume -c <profile_name>` 
   - Waterfox opens to AWS console with correct login, container name, and container icon from `~/.aws/config`

> _I have not tested this on linux or windows so I took my best guess on the path for windows based on [this reddit post](https://www.reddit.com/r/waterfox/comments/do1az7/waterfox_classic_201910_full_installer_the). If the path is incorrect the user can correct it as usual and/or file an issue to have the default updated._

### Potential risks
Leveraging some shared code with Firefox to launch / control Waterfox. If Waterfox diverges there could be issues here, however that seems unlikely since diverging may cause a lot of issues for Waterfox users in general. 

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -382,7 +382,7 @@ func AssumeCommand(c *cli.Context) error {
 			return err
 		}
 
-		if cfg.DefaultBrowser == browser.FirefoxKey || cfg.DefaultBrowser == browser.FirefoxStdoutKey {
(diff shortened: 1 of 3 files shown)


JoshuaWilkes (approved):
Thanks for the contribution!

Issue #155, opened 2022-04-22
[Feature Request] Support saving plaintext SSO tokens for interop with AWS SSM Session Manager plugin


chrnorm:
From the Common Fate Community Slack:

> We’re getting this error when trying to use aws ssm start-session - I couldn’t see anything online but now I’m reading Brian’s comment above is this something to do with the way granted is storing credentials? Is the SSM session manager plugin not compatible or something?
> ```
> ----------ERROR-------
> Encountered error while initiating handshake. KMSEncryption failed on client with status 2 error: Failed to process action KMSEncryption: Error calling KMS GenerateDataKey API: SSOProviderInvalidToken: the SSO session has expired or is invalid
> caused by: open /Users/kyle/.aws/sso/cache/<<long hex hash>>.json: no such file or directory
> ```

We should support a setting `UsePlaintextSSOTokens` in the Granted config file which would cause Granted to save the SSO tokens in plaintext so they are compatible with other AWS CLI plugins.

A longer term fix for this is to natively support AWS SSM Session Manager within Granted (perhaps by integrating with [gossm](https://github.com/gjbae1212/gossm).

underrun:
it would be nice to be able to use  profiles not just with other cli tools but with libraries that use aws SDKs which have a built in capability to use aws profile based creds in the same format as the CLI tool. gossm doesn't address this use case, but using the on disk SSO tokens that are created by `aws sso login` would.

lorengordon:
A profile that is using `credential_process` that calls `assume` to return the credentials may be another option to support this use case.

JoshuaWilkes:
Closing this as it has been implemented in #168 
Documentation on `assume --export` is here https://docs.commonfate.io/granted/usage/export-credentials

As a note, I think that issues with some AWS CLI tools is that the `assume` command will export AWS_PROFILE to the terminal environment.

In some CLI tools, where AWS_PROFILE is set in the environment, the credentials in the environment will be ignored and it will try to refetch the profile using the SDK.

I have found that in these cases, `unset AWS_PROFILE` in zsh or equivalent will make it use the assume credentials.

Another alternative as @lorengordon has pointed out is to use credential_process on the profile as described in this documentation https://docs.commonfate.io/granted/recipes/credential-process

chrnorm:
So the improvement made in #168 with `assume --export` does allow AWS session variables to be exported in plain text, however this issue relates to storing the AWS SSO token in plaintext.

[...]

PR #518, opened 2023-10-18
Allow access token to be exported to plain text on ~/.aws/sso/cache

cedieio:
### What changed?
This change would allow the access tokens to be exported to ~/.aws/sso/cache when the user run `assume --export`. Allowing `awscli` to use said token without the need to login again. Adjusted the keys for the file name as the default used is `sso_session_name`, `sso_start_url` is used if the session name is not present. Fix for issue #155.

### Why?
Allows `awscli` to use the access token without the need to login, and allows assume to use access token from `~/.aws/sso/cache` to avoid multiple logins

### How did you test it?
Ran `dassume --export` and tested via `aws s3 ls --profile test-profile` and checked `~/.aws/sso/cache` file location

Ran `dassume --export` with no `granted-aws-sso-tokens` on my keychain and valid access tokens on `~/.aws/sso/cache`,
checked the key-chain if `granted-aws-sso-tokens` is set, also tested in reverse(no plain-text token and has key-chain access-token)

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -524,6 +524,14 @@ func AssumeCommand(c *cli.Context) error {
 			}
 
 			clio.Successf("Exported credentials to ~/.aws/credentials file as %s successfully", profileName)
+
+			err := cfaws.ExportAccessTokenToCache(profile)
+
+			if err != nil {
+				return err
+			}
+
+			clio.Success("Exported access token to ~/.aws/sso/cache")
 		}
 
 		if execCfg != nil {
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -525,13 +525,16 @@ func AssumeCommand(c *cli.Context) error {
 
 			clio.Successf("Exported credentials to ~/.aws/credentials file as %s successfully", profileName)
 
+		}
+
+		if assumeFlags.Bool("export-sso-token") {
 			err := cfaws.ExportAccessTokenToCache(profile)
 
 			if err != nil {
 				return err
 			}
 
-			clio.Success("Exported access token to ~/.aws/sso/cache")
+			clio.Success("Exported sso token to ~/.aws/sso/cache")
 		}
 
 		if execCfg != nil {
--- a/pkg/cfaws/ssotoken.go
+++ b/pkg/cfaws/ssotoken.go
@@ -21,23 +21,23 @@ const (
 )
 
 type SSOPlainTextOut struct {
-	AccessToken string `json:"accessToken"`
-	ExpiresAt   string `json:"expiresAt"`
-	StartUrl    string `json:"startUrl"`
-	Region      string `json:"region"`
+	AccessToken    string `json:"accessToken"`
+	ExpiresAt      string `json:"expiresAt"`
+	SSOSessionName string `json:"ssoSessionName"`
+	StartUrl       string `json:"startUrl"`
+	Region         string `json:"region"`
 }
 
 // CreatePlainTextSSO is currently unused. In a future version of the Granted CLI,
 // we'll allow users to export a plaintext token from their keychain for compatibility
 // purposes with other AWS tools.
-//
-// see: https://github.com/common-fate/granted/issues/155
 func CreatePlainTextSSO(awsConfig config.SharedConfig, token *securestorage.SSOToken) *SSOPlainTextOut {
 	return &SSOPlainTextOut{
-		AccessToken: token.AccessToken,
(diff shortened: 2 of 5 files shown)


pkg/cfaws/ssotoken.go
```diff
@@ -47,26 +47,32 @@ func (s *SSOPlainTextOut) DumpToCacheDirectory() error {
 		return fmt.Errorf("unable to parse token to json with err %s", err)
 	}
 
-	err = dumpTokenFile(jsonOut, s.StartUrl)
+	// AWS uses the session name if present, else use startUrl
```
cedieio, line 50: https://docs.aws.amazon.com/cli/latest/userguide/sso-configure-profile-token.html
>The authentication token is cached to disk under the ~/.aws/sso/cache directory with a filename based on the session name.

pkg/assume/assume.go
```diff
@@ -524,6 +524,14 @@ func AssumeCommand(c *cli.Context) error {
 			}
 
 			clio.Successf("Exported credentials to ~/.aws/credentials file as %s successfully", profileName)
+
+			err := cfaws.ExportAccessTokenToCache(profile)
```
cedieio, line 528: Also not entirely sure if putting it inside `assumeFlags.Bool("export")` is good, open for suggestions.

JoshuaWilkes: Hi @cedieio 

reading through the use case for this I believe that making this a flag of the`granted sso login` command would be more suitable

for example `granted sso login --use-plain-text-cache` 

For convenience we could update the login command to list sso start URLs from the config file if they exist

cedieio: hmm, moving the flag to granted is a valid point, but could we keep `assume` reading from `~/.aws/sso/cache` bit ?

This would reduce number of logins when authenticating with `aws sso login --session test-session` 

This would ease up the flow on our part since we're heavily using AWS login and having to login using assume again is confusing for users.

cedieio: re: @JoshuaWilkes I just remembered a use case we encountered the other day for keeping the flag on assume.

we were using images that mount the entire `~/.aws` volume, some tools use the `~/.aws/credentials` and some use the `~/.aws/sso/cache` depending on the implementation.

For this our entry point is just assume so I guess for our use case we'll need assume to still export the creds to plaintext.

Moving this to granted would add extra steps for users that need the functionality if they use the assume binary only.

JoshuaWilkes: > re: @JoshuaWilkes I just remembered a use case we encountered the other day for keeping the flag on assume.
> 
> we were using images that mount the entire `~/.aws` volume, some tools use the `~/.aws/credentials` and some use the `~/.aws/sso/cache` depending on the implementation.
> 
> For this our entry point is just assume so I guess for our use case we'll need assume to still export the creds to plaintext.
> 
> Moving this to granted would add extra steps for users that need the functionality if they use the assume binary only.

Both Granted and Assume are actually bundled into the same binary `assumego` where granted is a symlink to `assumego`

`assume` is a shell script which calls `assumego` 

Not sure if this helps clarify your usecase

cedieio: Hi @JoshuaWilkes, thanks for getting back to me.

Just to clarify things if we move `--use-plain-text-cache` to `granted sso login` would it also work if we pass in `assume --use-plain-text-cache` ?

We're working with `assume` a lot and don't use `granted sso login` regularly as `assume` already does the stuff we need(logging in and setting cli AWS ENVs).

`assume` reading plain text from sso cache is also huge for us as it avoids duplicate logins when running `assume` if you have a valid sso access token in `~/.aws/sso/cache`

Cheers,
Cedric

cedieio: bumping this one @JoshuaWilkes

JoshuaWilkes: Hi Cedric!
So sorry that the review has dragged out so long.
I've had another read over it this morning and I think the only change I would like to make is to add a seperate flag to assume instead of bundling into --export
If you can add the flag --export-sso-token which exports to plain text just as you have it in the assume command.
The assumer code looks good to me.
I think it is best to separate the flags so that the behaviour doesn't change for existing users and so that for most users we maintain the added benefit of having the tokens in secure storage.
usage would be assume <my-profile> --export --export-sso-token

cedieio: Thanks for the feedback @JoshuaWilkes, I've separated the export sso token to it's separate flag.

Cheers,
Cedric

JoshuaWilkes (approved):
Thanks @cedieio!

PR #520, opened 2023-10-18
`-z` alias for `--export-all-env-vars`

dlambda:
### What changed?
`-z` alias for new `--export-all-env-vars` used to invoke previous ([sometimes unwanted](https://github.com/common-fate/granted/issues/263)) default behaviour.

### Why?
For less typing for my main use of `assume`

### Link to relevant docs PRs
https://github.com/common-fate/granted/pull/467#issuecomment-1767993740


chrnorm:
Thanks for this PR! 🙌  I think we'll also expose and document a config setting to revert the behaviour to the previous default if you prefer, as typing `-z` every time is just adding extra keystrokes and makes the CLI more cumbersome. I'm happy with `-z`, I think `-x` would be ideal but we already have a `-ex` flag to export credentials to `~/.aws/credentials` so this avoids confusion with `-ex`.

chrnorm (approved):

dlambda:
Great - thanks for this!  🤦  don't know why I messed up the branch & commit message, but did actually mean `-x`.

chrnorm:
🤦 oh my mistake! Given this has landed and been released we could consider eventually deprecating and eventually removing the `-ex` flag to prevent confusion, I don't think it's used that frequently

PR #542, opened 2023-11-03
Add support for gcp

meyerjrr:
### What changed?
Adds ability to use Granted with GCP to change contexts between projects. 
Open into the console for a particular project
Generate local config based off user access in GCP
Help with generating service account impersonated credentials through the --impersonate command

Refactor to the assume command to handle other cloud providers
### Why?
Make Granted accessible to users that may be using multi cloud

### Potential risks
Due to the refactor to the assume command, there is the risk that regressions could have been created in the process. I have tested out most of the usual assume functionality with AWS and have not ran into any regressions.

### Link to relevant docs PRs


meyerjrr:
Closing this due to being stale, might come back to this at some point but for now there were too many trade offs for it to be viable

PR #545, opened 2023-11-07
fix assume output for setting up SSO ENV VARS

### LINKED ISSUE #539: --active-role option broken for SSO
Since upgrading to a version with #467 in it, where the AWS ENV vars are not exported by default, I am unable to use the `--active-role` (or `-ar`) feature.

```
#-❯ assume -ar
[✘] one of the require environment variables was not found while loading an sso profile ['GRANTED_SSO_START_URL','GRANTED_SSO_REGION','GRANTED_SSO_ACCOUNT_ID','GRANTED_SSO_ROLE_NAME']
```

This is after I have done a successful `assume` where i can do other credential required things (such as `aws s3 ls`) and granted has put an `AWS_PROFILE` in my environment.

A snippet of my `~/.aws/confg` looks like this:

```
[profile my-profile]
granted_sso_start_url      = https://my-sso.awsapps.com/start
granted_sso_region         = my-region
granted_sso_account_id     = my-account
granted_sso_role_name      = my-role
common_fate_generated_from = aws-sso
credential_process         = granted credential-process --profile my-profile
```

I'm not entirely sure if there is some config I'm missing, or if I'm supposed to run something extra with the regular `assume` to make it export the `GRANTED_*` variables that `assume -ar` wants?

JoshuaWilkes:
### What changed?
Fixes #539 

When we made the breaking change to no longer export credentials for SSO profiles with credential process configured, there was a misconfiguration with the outputs which caused the special env var GRANTED_SSO to be set to "true"

This ENV var is used in conjunction with the --sso flags so that Granted knows that the profile doesn't exist in the config file but that it can reconstruct it virtually to reassume it.

### Link to relevant docs PRs


pkg/assume/assume.go
```diff
@@ -115,6 +115,8 @@ func AssumeCommand(c *cli.Context) error {
 	activeRoleProfile := assumeFlags.String("active-aws-profile")
 	activeRoleFlag := assumeFlags.Bool("active-role")
 
+	clio.Info(activeRoleProfile, activeRoleFlag, os.Environ())
```
meyerjrr, line 118: Im guessing this was a debugging statement? Can be removed now

meyerjrr (changes requested):

pkg/assume/assume.go
JoshuaWilkes: Yes good catch, this can be removed

PR #549, opened 2023-11-08
support exec using shell script for TTY applications

### LINKED ISSUE #250: --exec does not work for programs requiring a TTY [CLI-15]
Simple replication: `assume --exec 'vi test.txt'`

Solution doesn't appear to be trivial, but I think it would be good to (at least have an option to) use it with a TTY, so that interactive programs may be run.

JoshuaWilkes:
### What changed?
Fixes #250 

Uses the shell script to exec commands instead of Go, this enables TTY applications to work as expected

@TODO need to implement the exec feature in the other shell scripts

https://www.loom.com/share/dead201f196a44da8df631c562ff83ef

### Why?
Some programs like Bash would not work otherise

### Potential risks
Exec is now dependent on the shell script, so assumego cannot be called directly to exec. Though I don't think this would be an issue for anyone

Coudl mitigate this by still supporting Go based exec as well as shell based exec

### Link to relevant docs PRs


lgg42:
I'm so looking forward to this merge! In my case it is for using `terragrunt apply` wrapped by `assume`, e.g.

```
assume testing/RoleFoo --exec -- terraform apply
```

right now it is failing with
```
│ Error: error asking for approval: EOF
```

shwethaumashanker (approved):

PR #550, opened 2023-11-08
Output-sso-expiry-in-JSON

shwethaumashanker:
### What changed?
`granted sso-tokens expiry` returns a JSON output 

### Why?
It is easier to programmatically check if the token is expired 

### How did you test it?
Output of `dgranted sso-tokens expiry`. Also checked that you can run commands like `dgranted sso-tokens expiry | jq -r '[.[] | select(.is_expired == true)]'`

### Link to relevant docs PRs

### DIFF
--- a/pkg/granted/tokens.go
+++ b/pkg/granted/tokens.go
@@ -133,7 +133,7 @@ var TokenExpiryCommand = cli.Command{
 		if err != nil {
 			return err
 		}
-		clio.Logf(string(jsonData))
+		fmt.Println(string(jsonData))
 
 		return nil
 	},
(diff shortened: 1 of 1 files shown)


pkg/granted/tokens.go
```diff
@@ -111,8 +120,21 @@ var TokenExpiryCommand = cli.Command{
 				expiry = token.Expiry.Local().Format(time.RFC3339)
 			}
 
-			clio.Logf("%-*s (%s) expires at: %s", max, key, strings.Join(startUrlMap[key], ", "), expiry)
+			sso_expiry_data := sso_expiry{
```
chrnorm, line 123: @shwethaumashanker could you make it return JSON only if `--output=json` is set? 

I am thinking that we can use `--output=json` as a consistent flag in our CLIs to return JSON, otherwise we can print a more human-readable table output

chrnorm (changes requested):

pkg/granted/tokens.go
shwethaumashanker: I added `--json` flag since  `--output=json` was throwing a error: 
Screenshot 2023-11-08 at 5 27 41 PM

Let me know if this is okay or I can figure out a way to add `--output=json`

chrnorm (approved):

Issue #386, opened 2023-03-15
Registry sync should not delete profiles when failing to sync


Nevon:
Currently the way that the sync works is that it deletes all the content of the aws config that was created by granted, and then it fetches the new configuration and writes it back into the config file. If there is any error syncing a registry, all the other profiles are added into the config file.

I encountered this behavior the other day when I couldn't find any of the profiles I expected. The reason was that at the beginning of the workday I forgot to connect to the company VPN, which means that the first time I used `assume` it tried to sync my profiles but couldn't reach the remote repository and thus deleted most of my profiles from my config. I didn't notice this until later when I was very confused.

My suggestion would be that the configuration should have some kind of marker that shows which registry a profile was synced from. If syncing from a particular registry fails, print an error but don't delete that section.

Eddie023:
Hi @Nevon, thank you for raising this issue. Currently you can use `--prefix-all-profile` flag with `granted registry add` to add your registry name as prefix to show which registry a profile was synced from. 

On the registry sync failure case, removing the whole section on error was so that user will have visibility if there is some issues with the latest remote registry sync and can quickly update it. But in your case it was caused on the user end which is a bit of a problem. I will have a look on how we can make this experience better. Thanks  again for raising this!

webframp:
Just hit this myself:

```
[i] Syncing Profile Registries
[!] Sync failed for registry o11n
[Γ£ö] Completed syncing Profile Registries

[Γ£ÿ] Granted couldn't find any AWS profiles in your config file or your credentials file
[i] You can add profiles to your AWS config by following our guide:
[i] https://docs.commonfate.io/granted/getting-started#set-up-your-aws-profile-file
```

I had valid config of course, remote git repo is unreachable during sync (with no way to skip sync when running assume) so sync fails and config is deleted

chrnorm:
Fixed in #569

PR #569, opened 2023-12-07
Do not overwrite AWS config when sync fails

sosheskaz:
Fix issue #386 by writing the generated AWS config to a temp file, and replacing current AWS config if and only if all registry syncs succeed.

### What changed?

When `granted registry sync` fails, do not alter ~/.aws/config.

### Why?

Overwriting ~/.aws/config on failure creates a dangerous situation. If Github (or whatever VCS is being used) is down, and the sync runs automatically, then users are be unable to use the service.

Also, on machines that may not be connected to the VCS 24/7 (e.g. VPN, scheduled network outage, not connected to internet), granted registry sync may run automatically, leading to an inconvenience for a user because `granted registry sync` must be re-run manually.

### How did you test it?

1. `granted registry sync`
2. `wc < ~/.aws/config`
  1. `95488`
3. Disconnect network
4. `granted registry sync`
5. `wc < ~/.aws/config`
  1. With current version: `0`
  2. With this version (i.e. using `dgranted` instead of `granted`): `95488`

### Potential risks

This changes behavior which was previously intentional. However, I think that the downside of this is too high. Currently, we are effectively suppressing failures, and breaking the environment in the process. This will make the error louder, and more explicit.

There may be an argument that both failure modes should be supported. However, as a "Principle of Least Surprise" thing, I feel very strongly that this should be the default behavior. 

### Link to relevant docs PRs


Eddie023:
Thanks for the PR @sosheskaz. I do agree with your proposed change. However, this change can be a configurable option for backward compatibility and for people who would like the previous flow.

sosheskaz:
done

Eddie023 (approved):

chrnorm:
Thankyou @sosheskaz and thankyou @Eddie023 for the review 🙌 🙌

gautamg795:
FYI, per #600 it looks like this broke setups where `/tmp` is on a different device than `/home` (or wherever the user's AWS config might be stored)

It might be safer to open a temporary directory under the same path as the existing config file?

Issue #577, opened 2023-12-20
Values from `session` block does not get exported when using `--export-sso-token`


cedieio:
When using `session` on `~/.aws/config` the `sso_start_url` is not exported on the credentials under `~/.aws/sso/cache/`

e.g.
```
[sso-session test-session]
sso_region = ap-southeast-2
sso_registration_scopes = sso:account:access
sso_start_url = https://test/start#

[profile test-profile]
common_fate_generated_from = aws-sso
sso_account_id =000000000
sso_role_name = test-role
```

When running `assume test-profile --export-sso-token`

It gives the following credentials
```
{
   "accessToken"some_access_token",
   "expiresAt":"2023-12-20T19:31:16+10:00",
   "ssoSessionName":"test-session",
   "startUrl":"",
   "region":""
}
```

PR #578, opened 2023-12-20
Fix export lookup values when exporting credentials

cedieio:
### What changed?
When `session` is present on the config it should be where we get the `start_url` and `region`. Most configs will not have said config under `profile` if they have a session block. Session block takes priority for the configuration as it's considered as it's connecting profiles together.

Fix for #577

### Why?
Although it's working properly for `awscli` the generated token does not contain `sso_start_url` which could be problematic for other code that relies on that property when looking for a valid config.

### How did you test it?
`dassume --export-sso-token`

cat `~/.aws/sso/cache/<corresponding json>` should give out
```
{
   "accessToken"some_access_token",
   "expiresAt":"2023-12-20T19:31:16+10:00",
   "ssoSessionName":"test-session",
   "startUrl":"https://test.com",
   "region":"ap-southeast-2"
}
```

### Link to relevant docs PRs

### DIFF
--- a/pkg/cfaws/ssotoken.go
+++ b/pkg/cfaws/ssotoken.go
@@ -33,9 +33,12 @@ type SSOPlainTextOut struct {
 // purposes with other AWS tools.
 func CreatePlainTextSSO(awsConfig config.SharedConfig, token *securestorage.SSOToken) *SSOPlainTextOut {
 	ssoRegion := awsConfig.SSORegion
-	ssoStartURL := awsConfig.SSOStartURL
-	if awsConfig.SSOSession != nil {
+	if ssoRegion == "" && awsConfig.SSOSession != nil {
 		ssoRegion = awsConfig.SSOSession.SSORegion
+	}
+
+	ssoStartURL := awsConfig.SSOStartURL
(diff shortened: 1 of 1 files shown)


pkg/cfaws/ssotoken.go
```diff
@@ -32,12 +32,22 @@ type SSOPlainTextOut struct {
 // we'll allow users to export a plaintext token from their keychain for compatibility
 // purposes with other AWS tools.
 func CreatePlainTextSSO(awsConfig config.SharedConfig, token *securestorage.SSOToken) *SSOPlainTextOut {
+	ssoRegion := awsConfig.SSORegion
+	if ssoRegion == "" && awsConfig.SSOSession != nil {
```
cedieio, line 36: Was originally

```
if awsConfig.SSOSession != nil {
  ssoRegion = awsConfig.SSOSession.SSORegion
  ssoStartURL = awsConfig.SSOSession.SSOStartURL
}
```

Changed as profile can still have URLs set and the config still uses that

shwethaumashanker (approved):

Issue #388, opened 2023-03-20
`granted credentials rotate` does not delete the former credentials


matthewhembree:
When running:

```
➜ granted credentials rotate
? Profile Name: default
[✔] Access Key of 'default' profile has been successfully rotated and updated in secure storage

➜ granted credentials rotate default
? Profile Name: default
[✘] operation error IAM: CreateAccessKey, https response error StatusCode: 409, RequestID: REDACTED, LimitExceeded: Cannot exceed quota for AccessKeysPerUser: 2
```

The former credentials are deactivated, but not deleted.

Thanks!

edit: wrong output. updated to be correct.

n3s7or:
That feature was developed following AWS guidelines for [access key rotation](https://aws.amazon.com/blogs/security/how-to-rotate-access-keys-for-iam-users/), i.e. checking if your applications still working after updating to the new key, and if it doesn't you can rollback to the previous key. So if the previous access key was deleted there's no rolling back.

It's an easy change, just calling delete after [this line](https://github.com/common-fate/granted/blob/91242aa665a3614ffc9698ac8886b2ebaebd1f9b/pkg/granted/credentials.go#L577)

montesuma:
It would be nice to have a configuration option for this. I think many people use aws key rotation on schedule via cron task

PR #582, opened 2023-12-22
Delete former credentials when rotating (`granted credentials rotate`)

n3s7or:
### What changed?
Added code to delete old credentials.

### How did you test it?
I rotated my own AWS credentials

No

### Link to relevant docs PRs

### DIFF
--- a/pkg/granted/credentials.go
+++ b/pkg/granted/credentials.go
@@ -583,6 +583,11 @@ var RotateCredentialsCommand = cli.Command{
 			return err
 		}
 
+		_, err = iamClient.DeleteAccessKey(c.Context, &iam.DeleteAccessKeyInput{AccessKeyId: &t.AccessKeyID})
+		if err != nil {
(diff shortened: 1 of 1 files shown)


chrnorm (approved):
Thankyou @n3s7or!!

chrnorm (changes requested):
Actually, just looking through the PR and original issue #388 and it would be good to introduce this as opt-in behaviour. Could you please add a flag `--delete` to the command, and then only delete the key if the flag is set?

You can do so by adding
```go
&cli.BoolFlag{Name: "delete", Usage: "delete the previous active key"}
```
to the command definition and then using

```go
if c.Bool("delete") { }
```

to check if the flag is set.

n3s7or:
Thanks for your input @chrnorm
The change is done :)

chrnorm (approved):
🙌 thankyou @n3s7or!

PR #616, opened 2024-03-01
Add support for refreshable AWS SSO tokens

### LINKED ISSUE #615: Support refreshable AWS SSO sessions
Granted currently uses the legacy AWS SSO login flow. This is the current login screen that you see, which indicates the legacy flow:

image

The legacy login flow returns non-refreshable access tokens with a fixed 8hr expiry.

The new AWS SSO login flow looks like this:

image

It returns shorter-lived access tokens with a 1hr expiry, and returns refresh tokens allowing the access token to be refreshed. An organization's administrator can [configure the session duration](https://docs.aws.amazon.com/singlesignon/latest/userguide/configure-user-session.html) up to a maximum of 90 days.

chrnorm:
### What changed?

This PR adds opt-in support for the new IAM Identity Center refreshable token login flow. You can opt-in to using refreshable tokens by configuring your AWS profiles to include `granted_sso_registration_scopes` as follows:

```diff
[profile example]
granted_sso_account_id = 123456789012
granted_sso_role_name = AWSAdministratorAccess
+ granted_sso_registration_scopes = sso:account:access
granted_sso_start_url = https://commonfate.awsapps.com/start
granted_sso_region = ap-southeast-2
credential_process         = granted credential-process --profile example
```

At the moment, support for SSO registration scopes has been implemented only for profiles using the [Granted Credential Process](https://docs.commonfate.io/granted/recipes/credential-process) configuration.

Some background here; the native AWS CLI way of configuring this is by using the `sso-session` entry as follows:

```
[profile AWSAdministratorAccess-123456789012]
sso_session = commonfate
sso_account_id = 123456789012
sso_role_name = AWSAdministratorAccess
region = ap-southeast-2

[sso-session commonfate]
sso_start_url = https://example.awsapps.com/start
sso_region = ap-southeast-2
sso_registration_scopes = sso:account:access
```

I would like Granted to support this configuration too, however the AWS v2 Go SDK does not yet support reading `sso_registration_scopes`. So the initial implementation here just uses `granted_sso_registration_scopes`.

In future it's likely we'll make the refreshable token flow the default, but for now this behaviour is opt-in.

Additionally, this PR tidies up some of our AWS SSO login flow code. In the Granted codebase the `cfaws` package has grown to be quite large. I've moved the SSO login flow code _out_ of this package into a new `idclogin` package.

### Why?
Fixes #615.

### How did you test it?
Tested and verified on MacOS by manually logging in a few times to verify the functionality was correct. I tested the refresh flow by manipulating the `Expiry` value of the token inside my keychain.

I also tested the case where the refresh token is invalid, by overriding the `refreshToken` value inside my keychain and overriding the `Expiry` to force a refresh. The behaviour here is that an error is logged to stderr and the user is prompted to reauthenticate, which is what we want.

Given that no platform-specific keychain functionality has been modified I expect this to work across other OSes too.

### Potential risks
- May impact SSO login flow


go.mod
```diff
@@ -18,9 +18,11 @@ require (
 )
 
 require (
+	github.com/alessio/shellescape v1.4.2
```
chrnorm, line 21: I ran `go mod tidy` and cleaned up `go.mod` and `go.sum`

pkg/securestorage/sso_token_storage.go
```diff
@@ -20,36 +23,96 @@ func NewSecureSSOTokenStorage() SSOTokensSecureStorage {
...
+
+	return &newToken
 }
 
 // Attempts to store the token, any errors will be logged to debug logging
 func (s *SSOTokensSecureStorage) StoreSSOToken(profileKey string, ssoTokenValue SSOToken) {
 	err := s.SecureStorage.Store(profileKey, ssoTokenValue)
 	if err != nil {
-		clio.Debugf("%s\n", errors.Wrap(err, "writing sso token to credentials cache").Error())
+		clio.Debugf("writing sso token to credentials cache: %s", err.Error())
```
chrnorm, line 108: I've fixed the syntax up here - the previous syntax was difficult to read

shwethaumashanker (approved):

PR #618, opened 2024-03-06
Revert breaking change to 'granted login' with additional prompt for SSO scopes

chrnorm:
This fixes an issue where 'granted login' would trigger an interactive prompt if --sso-scopes were not provided. This was 
an unintentional breaking change caused by #616.

This is an issue if you have written shell scripts that call `granted login`, because unless `--sso-scopes` is provided, the command will now hang and wait for user input.


shwethaumashanker (approved):
Tested with bash script calling `granted login`

PR #626, opened 2024-03-22
feat: Use aws:username for IAM initiated federated console sessions.

### LINKED ISSUE #625: IAM Federated logins (console) should have easily attributable username in Cloudtrail list view.
When using federated login with IAM credentials, the Federated token username is not easily attributable to a real user in the Cloudtrail event logs.

For instance, on the CloudTrail event entry for GetFederationTokenInfo, the user name might be John.Smith, but the created federated token user's user name would be something like AIDA... as currently the `aws:userid` ([ref1](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_variables.html#principaltable) [ref2](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_identifiers.html#identifiers-unique-ids)) is being used. 

This makes it hard to quickly attribute the real user to the federated user's events from the list view.

image

This is different from the assumed-role/SSO types of logins where the user name appears to correctly propagate into the session user (e.g. `John.Smith@company.com`).

The proposal is to use the `aws:username`, which would be something like John.Smith.

image

This is to further enhance the work done in #469 by addressing the feedback in the comments.

Originally referenced in #387 

Thanks!

matthewhembree:
### What changed?

Closes #625 

The changes refactor the way federation token ID is used for AWS IAM credentials. Instead of relying on the userID which was previously parsed, the code now uses the userName which is more easily attributable to the IAM user name in the Cloudtrail events list view.

Old:

image

New:

image

### Why?

In the Cloudtrail console's event history view, the IAM user name will now display in the `user name` column. Previously, the `user id` would display (e.g. AIDA.....).

### How did you test it?

1. Add IAM credential with `granted credentials add`.
2. Opened a console session with `assume -c`.
3. Created an S3 bucket through the S3 console.

### Link to relevant docs PRs

### DIFF
--- a/pkg/cfaws/assumer_aws_iam.go
+++ b/pkg/cfaws/assumer_aws_iam.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
 	"github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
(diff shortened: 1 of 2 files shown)


chrnorm (approved):
🙌 Thanks very much and congratulations here on your first contribution to Granted @matthewhembree! This looks great, I've added an additional unit test around the behaviour here https://github.com/common-fate/granted/pull/626/commits/3e5ea5f6a7c3ec6bdc34c4b9c964a6fa63d9caed

PR #631, opened 2024-04-04
fix bug where every line is removed in config

Nepoxx:
### What changed?

The function `RemoveLine` will no longer remove all lines.

### Why?

`removeLine := strings.Contains(line, line)` would always be true and thus all lines would be removed.


chrnorm (approved):
Thankyou @Nepoxx!

PR #635, opened 2024-04-16
Ensure "firefoxstdout" doesn't get prefix-matched as "firefox"

gautamg795:
### What changed?
The `GetBrowserKey` function has been fixed to return the correct `FirefoxStdoutKey`.

### Why?
Previously, this value would never be returned as the function performs a `Contains` match, and `firefoxstdout` does contain the string "firefox" which would return the wrong key. 

### How did you test it?
Honestly, just visual inspection after finding that providing `FirefoxStdout` to `granted browser set` was still prompting for a path.

### Potential risks
None


chrnorm (approved):
Thankyou @gautamg795! 🙌

PR #636, opened 2024-04-16
Add apigateway to service map

treuherz:
### What changed?

Add a service map entry for AWS API Gateway (`apigateway`)


chrnorm (approved):
Thankyou @treuherz! 🙌

PR #640, opened 2024-04-19
Increase backoff for attempted assume

meyerjrr:
### What changed?
Implemented a fibonacci backoff retry when requesting access via Common Fate and waiting for the access to be live. 

### Why?
The current backoff retry was too short and caused some users to get bad UX when requesting access with Common Fate and subsequentially assuming the role. The access would not be provisioned yet and would receive the error
```
[✘] no access: operation error SSO: GetRoleCredentials, https response error StatusCode: 403, RequestID: 123456-db64-42f3-88a8-asdfjk3434, api error ForbiddenException: No access
```

### How did you test it?
Assuming a role with the live assume command with the current functionality and replicating the error above.
The assuming the same role via the dev assume command dassume and monitored the result. The wait was longer and no error was returned

### Potential risks
_For users that are not using Common Fate and try assuming a role they do not have access to will have to wait for the retry_
- I have tested the above and this is not affected

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -1,6 +1,7 @@
 package assume
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/url"
@@ -33,6 +34,7 @@ import (
 	cfflags "github.com/common-fate/granted/pkg/urfav_overrides"
 	"github.com/fatih/color"
 	"github.com/hako/durafmt"
+	sethRetry "github.com/sethvargo/go-retry"
 	"github.com/urfave/cli/v2"
 	"gopkg.in/ini.v1"
 )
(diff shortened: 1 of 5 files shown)


pkg/assume/assume.go
```diff
@@ -505,12 +507,20 @@ func AssumeCommand(c *cli.Context) error {
...
 			if retry {
-				for i := 0; i < 5; i++ {
+
+				b := sethRetry.NewFibonacci(time.Second)
+				b = sethRetry.WithMaxDuration(time.Minute*1, b)
+				err = sethRetry.Do(c.Context, b, func(ctx context.Context) (err error) {
 					creds, err = profile.AssumeTerminal(c.Context, configOpts)
 					if err == nil {
-						break
+						return sethRetry.RetryableError(err)
```
chrnorm, line 516: I won't block merge on this but a couple of extra thoughts @meyerjrr:

- retrying on all errors could cause `assume` to feel as though it's hanging in the event of a network error, we could improve this in future by checking the error type here
- it would be good to print to the user something like `trying to assume the role... [time elapsed = 10s]` similar to Terraform so that they know we are retrying

PR #643, opened 2024-04-25
CLI flags to specify storage and profile to clear credential cache without prompt

jsproede:
### What changed?
Added CLI flags to `ClearCommand` to be able to clear specific profile credential cache directly without any prompt.

### Why?
We're working in an AWS environment where we are limited to an 1 hour maximum session duration. Without clearing the credential cache, we're not able to renew the session as the cached credentials are always used. We have defined certain Terminal aliases to quickly switch between different AWS profiles, therefore it would be really helpful, to be able to clear the credential cache within our aliases 🙂 

### How did you test it?
* `granted cache clear` without any CLI flags defaults to ask the user for the storage and profile, no changes here.
* `granted cache clear --storage session-credentials --profile [profile_name]` skips the prompts and clears the cache entry immediately.

### Potential risks
None.

It does not fix any bug, but it is no breaking change.

### DIFF
--- a/pkg/granted/cache.go
+++ b/pkg/granted/cache.go
@@ -56,17 +56,24 @@ var ListCommand = cli.Command{
 var ClearCommand = cli.Command{
 	Name:  "clear",
 	Usage: "Clear cached credential from the secure storage",
+	Flags: []cli.Flag{
+		&cli.StringFlag{Name: "storage", Usage: "Specify the storage type"},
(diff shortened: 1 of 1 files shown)


chrnorm (approved):
This was sorely needed, thanks for the contribution. I've been considering adding `granted cache clear --all` too to nuke the entire local cache.

PR #649, opened 2024-05-01
improve logging when retrieving the federated console URL fails

chrnorm:
### What changed?
Adds extra logging of the response body when opening the console fails

### Why?
We're currently seeing reports of 'opening console failed with code 400' but this is difficult to debug, because the HTTP body is not printed.

### How did you test it?
Untested but will have a community member run a dev build of this, ideally.

### Potential risks
Low to none - just adds extra body logging in the error case.

Yes

### Link to relevant docs PRs
N/A

### DIFF
--- a/pkg/console/aws.go
+++ b/pkg/console/aws.go
@@ -3,6 +3,7 @@ package console
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


chrnorm:
Closing as this did not aid in solving the issue

Issue #650, opened 2024-05-02
For granted sso generate/populate commands add support to persist preferences in .granted/config


ibottamike:
The granted sso generate/populate commands have input for StartURL, SsoRegion, Prefix, Source, NoCredentialProcess, and ProfileTemplate.

Only having the parameters available as input variables require them to be enter every time the commands are run. In a dynamic environment where AWS accounts are added/removed, this requires the variables to be stored outside of granted configuration.  

Add the capability to store the configuration values in `.granted/config` to be consistent with `registry add/sync` and assume browser preferences.

chrnorm:
Implemented in #651!

PR #651, opened 2024-05-02
for granted sso command, add storing parameters in input config https://github.com/common-fate/granted/issues/650

mikesarver:
### What changed?

For `granted sso generate/populate` commands add support to persist preferences in .granted/config

### Why?
[issues/650](https://github.com/common-fate/granted/issues/650)
Persists `granted sso` preferences in `~/.granted/config`. This allows storing preferences in granted config across commands.  This allows persisting the data once in config and used across when running commands to sync profiles.

### How did you test it?
Tested manually.

### Link to relevant docs PRs


pkg/config/config.go
```diff
@@ -55,6 +55,16 @@ type Config struct {
 		Variables               map[string]string `toml:",omitempty"`
 		Registries              []Registry        `toml:",omitempty"`
 	} `toml:",omitempty"`
+
+	Sso struct {
```
chrnorm, line 59: Rather than a single 'Sso' struct, I am thinking we could make this a map to allow us to support multiple AWS SSO configurations (for example, if a user has multiple AWS organizations).

e.g.
```go
type Config struct {
  ...
  SSO map[string]AWSSSOConfiguration
}

type AWSSSOConfiguration struct {
  StartURL string
  SSORegion string
  Prefix string
  ...
}
```

In the config file, this would then look like:
```toml

[SSO.default]
StartURL = "...."
SSORegion = "us-east-1"
...
```

In the case where a user only has *one* `AWSSSOConfiguration` we would use the current behaviour (which would solve your use case in the same way as your current implementation). In the case where they have two or more though, they could specify a particular one with `--config` (or a similar flag).

pkg/granted/sso.go
```diff
@@ -155,13 +179,14 @@ var PopulateCommand = cli.Command{
...
-			NoCredentialProcess: c.Bool("no-credential-process"),
-			Prefix:              c.String("prefix"),
+			ProfileNameTemplate: profileNameTemplate,
+			NoCredentialProcess: noCredentialProcess,
+			Prefix:              prefix,
 			PruneStartURLs:      pruneStartURLs,
 		}
 
-		for _, s := range c.StringSlice("source") {
+		sources := coalesce.StringSlice(c.StringSlice("source"), cfg.Sso.Source)
```
chrnorm, line 188: nit: could you please vendor this function into the current package so that we can avoid the additional package dependency? (the same as what you have done with `coalesceString()`)

pkg/config/config.go
```diff
@@ -55,6 +55,16 @@ type Config struct {
...
+		// add any global configuration to sso profile commands defaults here.
+		StartURL            string
+		SsoRegion           string
```
chrnorm, line 62: nit: we prefer to capitalise acronyms in our Go code, so 'SsoRegion' would be 'SSORegion'

chrnorm (changes requested):
Thanks very much for this @mikesarver, I am very keen to improve the UX with profile generation and avoiding having to pass all those extra flags. I've requested changes on a couple of smaller nits, as well as some thinking around how we might make this useable for folks who have multiple AWS SSO instances they work with.

pkg/granted/sso.go
ibottamike: Another thought I had after submitting, should the command line append vs replace the list. I don't use sources, so would love your input.
`sources := append(c.StringSlice("source"), cfg.Sso.Source...)`

chrnorm: Great question, some context on this is that we added `sources` to allow [Glide](https://github.com/common-fate/glide) users to call this command (`--source commonfate` would use Glide). To my knowledge it isn't widely used and we should deprecate `--source commonfate` because Glide is being sunset. This leaves `--source aws-sso` as the only remaining source flag.

Upon some further thought, what I think we should do here is remove the `Source` field that you've added to the SSO config for now:

```diff
Sso struct {
		// add any global configuration to sso profile commands defaults here.
		StartURL            string
		SsoRegion           string
		Prefix              string
-		Source              []string
		NoCredentialProcess bool
		ProfileTemplate     string
	}
```

We'll deprecate the `--source` flag and default to using AWS SSO.

In future if we add additional sources again, we can re-introduce it in the config (and this will be a non-breaking change, as the default source will remain to be AWS SSO).

chrnorm: It's unlikely that we'll add additional custom sources in future here too -- we're planning on extending the [Profile Registry API](https://docs.commonfate.io/granted/usage/profile-registry) to be more flexible, rather than using this command.

ibottamike: Thank you for the detail explanation.  I will keep a lookout for the extended profile registry features.  We are planning on using this heavily.

pkg/granted/sso.go
```diff
@@ -47,6 +48,7 @@ var GenerateCommand = cli.Command{
 	Usage:     "Prints an AWS configuration file to stdout with profiles from accounts and roles available in AWS SSO",
 	UsageText: "granted [global options] sso generate [command options] [sso-start-url]",
 	Flags: []cli.Flag{
+		&cli.StringFlag{Name: "config", Usage: "Specify the SSO config section ([SSO.name])", Value: "default"},
```
chrnorm, line 51:
```suggestion
		&cli.StringFlag{Name: "config", Usage: "Specify the SSO config section in the Granted config file ([SSO.name])", Value: "default"},
```

chrnorm (approved):
Tested this locally and it's working great! Thanks so much @mikesarver 🙌

PR #653, opened 2024-05-06
add command for setting up new cf registry

meyerjrr:
### What changed?
- Adds functionality for adding a second type of profile registry with Granted. The source being Common Fate. When added will work the same as the github registry in that it will sync profiles from Common Fate but also based on the users access level.

### Why?
Common Fate users will now be able to have the config auto filled with account:role combinations they have access to. 

### How did you test it?
Pulled profiles from a test Common Fate deployment.
Tested old git registry functionality to confirm that there were no regressions 

### Potential risks
Git registry functionality stopping working

### Link to relevant docs PRs


pkg/granted/registry/add.go
```diff
@@ -76,84 +77,163 @@ var AddCommand = cli.Command{
...
 
-		registry, err := gitregistry.New(gitregistry.Opts{
-			Name:         name,
-			URL:          gitURL,
-			Path:         pathFlag,
-			Filename:     configFileName,
-			RequiredKeys: requiredKey,
-			Interactive:  true,
-		})
+		if IsGitRepository(URL) {
```
chrnorm, line 80: The `IsGitRepository` check here is a bit brittle - it's possible that someone could be self-hosting a git repo at a URL which isn't GitLab or GitHub (or even simply using Azure DevOps, or Bitbucket etc).

Rather than trying to infer if it's a git repo or using our RPC protocol automatically, let's add a `--type` flag for users to indicate which type of registry they want to add.

e.g.
```
# add a HTTP registry using the granted.registry.aws.v1alpha1 protocol
granted registry add --type http --url https://commonfate.example.com

# add a git registry
granted registry add --type git --url https://some.git.repository.url
```

We should default to `--type=git` to preserve the current behaviour.

pkg/granted/registry/add.go
```diff
@@ -76,84 +77,163 @@ var AddCommand = cli.Command{
...
+			}
 
-			options := []string{DUPLICATE, ABORT}
+			return nil
+		} else {
 
-			in := survey.Select{Message: "Please select which option would you like to choose to resolve: ", Options: options}
-			var selected string
-			err = testable.AskOne(&in, &selected)
+			registry, err := cfregistry.New(cfregistry.Opts{
```
chrnorm, line 161: Why did the code in this section onwards need to be changed? (on initial glance it looks mostly the same as the current implementation, but just shifted around?)

pkg/granted/registry/cfregistry/cfregistry.go
```diff
@@ -0,0 +1,93 @@
...
+func New(opts Opts) (*Registry, error) {
+
+	cfg, err := config.LoadDefault(context.Background())
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.APIURL != opts.URL {
+		return nil, fmt.Errorf("passed url does not match url in Common Fate (cf) config. active context API URL is: %s", cfg.APIURL)
+	}
```
chrnorm, line 34: Let's remove this check. One of the goals of implementing this in Granted is that teams should be able to implement their own HTTP-based registry, by implementing the `granted.registry.aws.v1alpha1` protobuf API (for example, a team could write a custom Lambda function to serve AWS profiles). 

This check forces people to use the Common Fate API, which shouldn't be required for this feature

```suggestion
```

pkg/granted/registry/registry.go
```diff
@@ -18,6 +20,12 @@ type loadedRegistry struct {
 	Registry Registry
 }
 
+func IsGitRepository(url string) bool {
+	regex := regexp.MustCompile(`(?:https:\/\/github\.com\/|git@github\.com:)[a-zA-Z0-9_-]+\/[a-zA-Z0-9_-]+(?:\.git)?`)
+
+	return regex.MatchString(url)
```
chrnorm, line 26: as per above, let's remove this in favour of a `--type` flag

chrnorm (changes requested):

pkg/granted/registry/add.go
meyerjrr: The only thing changed in this file was including the check at the top to handle the registry type. I'd say the diff is effected by this as it is now wrapped in a conditional

pkg/granted/registry/cfregistry/cfregistry.go
meyerjrr: @chrnorm Ahh yep this was different to my understanding. In that case we will need to handle API authentication better here too. 

At the moment it will rely on the Granted user having set up cf and having context already existing in the config and pulling from that to successfully authenticate with the CF API endpoint.
In our case we are setting up a client handler which calls Common Fate with the oidc auth. Same as how the `cf` cli would.  

Another case which you have mentioned is via a lambda call. This could be served directly via the lambda or via an api gateway, either way the auth here would be with AWS IAM. 

Either way the logic for handling this will differ for every case that an end user will need to use. Unless the API has no auth attached. Is the solution here to make the http registry generic enough to just handle a bearer token to be attached to an api call, leaving adding the logic for obtaining that auth token to the end user?

chrnorm: I am thinking here for the initial implementation let's leave the existing auth as-is using OIDC. Because we use OIDC with Common Fate, if a user wanted to set up a custom Granted Profile Registry service they could also use OIDC to authenticate.

In that case the user would *still* use the `~/.cf/config` file to specify the OIDC issuer, even if they set up a custom HTTP registry.

pkg/granted/registry/add.go
```diff
@@ -33,8 +34,10 @@ var AddCommand = cli.Command{
 		&cli.BoolFlag{Name: "prefix-all-profiles", Aliases: []string{"pap"}, Usage: "Provide this flag if you want to append registry name to all profiles"},
 		&cli.BoolFlag{Name: "prefix-duplicate-profiles", Aliases: []string{"pdp"}, Usage: "Provide this flag if you want to append registry name to duplicate profiles"},
 		&cli.BoolFlag{Name: "write-on-sync-failure", Aliases: []string{"wosf"}, Usage: "Always overwrite AWS config, even if sync fails (DEPRECATED)"},
-		&cli.StringSliceFlag{Name: "required-key", Aliases: []string{"r", "requiredKey"}, Usage: "Used to bypass the prompt or override user specific values"}},
-	ArgsUsage: "--name <registry_name> --url <repository_url>",
+		&cli.StringSliceFlag{Name: "required-key", Aliases: []string{"r", "requiredKey"}, Usage: "Used to bypass the prompt or override user specific values"},
+		&cli.StringFlag{Name: "type", Aliases: []string{"t"}, DefaultText: "git", Usage: "specify the type of granted registry source you want to set up. Default: git"}},
```
chrnorm, line 38: I think `--type` is fine (I'm hesitant to add aliases too aggressively because `-t` becomes reserved forever for it, I prefer for us to introduce them *only* for very frequently used commands initially)

```suggestion
		&cli.StringFlag{Name: "type", DefaultText: "git", Usage: "specify the type of granted registry source you want to set up. Default: git"}},
```

chrnorm (changes requested):
Looks good, minor change requested. Aside from the code change we also need to update `granted login` to call the new OIDC login flow - otherwise a user who has not installed `cf` will not be able to authenticate to the registry.

chrnorm (approved):

PR #659, opened 2024-05-15
fix profile registry sync for registries without a type

chrnorm:
they should default to 'git'


shwethaumashanker:
Screenshot 2024-05-15 at 12 32 36 PM
Tested locally

Issue #614, opened 2024-02-29
Assumed Role - Randomized UserId gntd-<random-suffix>


Siushaga:
I'm facing a similar issue to why PR's were implemented, such as the ones below:
https://github.com/common-fate/granted/pull/189
https://github.com/common-fate/granted/pull/469

The difference for our use case is that we use a GlobalReadOnly role in a central account that has the ability to assume a ReadOnly role in any of the AWS accounts that we own. The GlobalReadOnly (or any "single hop" role) is presenting the UserId as our Active Directory UserId, as expected. However, if a role assumption is introduced then the session name is receiving a name like "gntd-<random-suffix>".

This layer of anonymity in CloudTrail where we can't attribute to what individual user did something will render this utility unusable for us, so I'm hopeful this is something that can be addressed. 

In my ~/.aws/config file I have a profile for global-read-only that uses a credential_process to call okta-cli to process-credentials from the central account's Okta IdP and GlobalReadOnly role, while all other profiles set the role_arn to the ReadOnly role in each account along with a source_profile set to the global-read-only profile.

Doing an assume on one of the cross account read only roles works fine to get console access or temporary creds in my terminal. However, `aws sts get-caller-identity` shows like this
`{
  "UserId": "REDACTED_PRINCIPAL_ID:gtnd-2d1LCv2dD5bIlx0hzQl4gpaw2pm",
  "Account": "REDACTED",
  "Arn": "arn:aws:sts::REDACTED:assumed-role/ReadOnly/gtnd-2d1LCv2dD5bIlx0hzQl4gpaw2pm"
}`

Do an assume on the global-read-only profile, or any of my other more permissive profiles that don't involve cross account access and it looks correct with AD user identity.
`{
  "UserId": "REDACTED_PRINCIPAL_ID:AD_USER_ID",
  "Account": "REDACTED",
  "Arn": "arn:aws:sts::REDACTED:assumed-role/GlobalReadOnly/AD_USER_ID"
}`

We thought perhaps we needed to include `sts:TagSession` or `sts:TransitiveTagKeys` on the role's permissions, but I'm still seeing the same issues.

PR #665, opened 2024-05-22
feature/use-aws-user-id-sso

omaotzu:
Firstly to caveat this is my first time contributing to a project and also my first time writing anything in go so please bear with me. 

My knowledge isn't great enough to foresee any untoward complications but we ran into several issues related to the randomised gtnd user id and thought I would give it a crack.

Any guidance or feedback would be greatly appreciated and understand that more work may be required on my part (or you can just outright dismiss this 😆 if it's stupid or my approach is wildly off)

Primarily the issues that we faced were the same as outlined within this issue: https://github.com/common-fate/granted/issues/614

### What changed?

- Optionally call stsClient.GetCallerIdentity when assuming to fetch the user id attach that via the role session name
- Provide a config flag (uu) so this can optionally be used within the cli
- Called Getenv to check for the env file flag to permanently within global environment variables

### Why?

- Firstly the random gtnd user id effectively made cloudtrail tracing by user physically impossible, each individual session had its own id and it was very difficult to attribute individual user actions to a specific user. This posed security concerns for us as we couldn;t  use global deny all policies on these roles as there was no user specific attribute to add into the iam.
- Some things within the aws console are specific to the user the created them. A few examples of this live within the lambda console itself. Private saved events for triggering are unique to the user and thus only exist for the lifespan on the unique gtnd user id. Some of our data team also wish to manually edit lambda function code in console within non production accounts to quickly iteration on solutions, but these are also unique to the specific user.

Thus finding a method of attributing a userId that was unique to the user rather to the session is imperative for us to continue using this package.

### How did you test it?

- Building the package locally with the proposed changes and adding in the config flag followed by running sts's get caller identity outputted the the random gtnd id without the flag and my specific user with the flag
- Adding in `GRANTED_USE_USER_ID=true` within my zshrc defaulted this to always specific user when I assumed
- Assuming into a variety of our accounts and checking cloudtrail in each of these environments allowed my to search specifically for my user id
- Adding in an explicit deny all policy to a role I had actively assumed that was specifically attributed to my user id immediately blocked all permissions within aws

### Potential risks
- Error catching
- Splitting of the called.UserId by `:` rather than a more thorough approach
- Unsure and more feedback would be greatly appreciated!

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -288,6 +288,7 @@ func AssumeCommand(c *cli.Context) error {
 		MFATokenCode: assumeFlags.String("mfa-token"),
 		Args:         assumeFlags.StringSlice("pass-through"),
 		DisableCache: assumeFlags.Bool("no-cache"),
+		UseUserId:    assumeFlags.Bool("use-user-id"),
 	}
 
 	// attempt to get session duration from profile
--- a/pkg/assume/entrypoint.go
+++ b/pkg/assume/entrypoint.go
@@ -56,6 +56,7 @@ func GlobalFlags() []cli.Flag {
 		&cli.BoolFlag{Name: "wait", Usage: "When using Granted with Common Fate the assume will halt while waiting for the access request to be approved."},
 		&cli.BoolFlag{Name: "no-cache", Usage: "Disables caching of session credentials and forces a refresh", EnvVars: []string{"GRANTED_NO_CACHE"}},
 		&cli.StringSliceFlag{Name: "browser-launch-template-arg", Usage: "Additional arguments to provide to the browser launch template command in key=value format, e.g. '--browser-launch-template-arg foo=bar"},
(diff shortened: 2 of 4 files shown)


chrnorm:
Hi @omaotzu, thanks so much for opening a contribution here. I certainly agree that the current random IDs make attribution difficult in CloudTrail and would like to improve this.

One concern I have is that the `RoleSessionName` field is user-controlled. If you rely on this for detection, it is possible for an attacker to manipulate this field and change the user ID that events are attributed to. Are there any IAM conditions that we'd be able to apply here to ensure users can only apply a `RoleSessionName` that matches claims from the IAM Identity Center SAML provider?

chrnorm:
cleaning up some old PRs, sorry for not getting this merged :( I think we're still waiting on a way for AWS to prevent user manipulation of session names if we take an approach like this. Would welcome a PR in future to fix this!

PR #672, opened 2024-06-03
update error messages and auto login when grant token is expired

meyerjrr:
### What changed?
Adds a case when using `assume` with Common Fate to handle automatically trying to login in when token has expired.

Also includes an improvement to the error message to be more informational on next steps to fix the error.

### Why?
Some users were getting errors when using the integration which required prior set up with the cf cli. This has now been included in Granted. Extra information given in error messages to assist in the process.

### Link to relevant docs PRs


chrnorm (approved):
I've made some minor tweaks on top of this to programmatically check the error returned by `config.LoadDefault()`

PR #679, opened 2024-06-05
Credential process prompts to use 'granted request latest' when possible

JoshuaWilkes:
### What changed?
Adds a new command `granted request latest` which is used in conjunction with the credential process and Common Fate.
Now, when you don't have access to a role in AWS and you are using granted, you will be prompted to request access to it.

Also fixes some bugs with the NoAccess hook where the reason prompt was blocked by the spinner when the --reason flag was not used, this would leave the cli hanging waiting for an input.

Pass through the --confirm prompt to better support non interactive usecases

### Link to relevant docs PRs


JoshuaWilkes:
I accidentally added some extra commits which I have reverted

PR #682, opened 2024-06-12
add flag that waits for access to be active

meyerjrr:
### What changed?
New flag added to the assume command `--wait` which when used with a Common Fate integration, will wait for 15 minutes reattempting to grab credentials. 

### Why?
This flag can be used when requesting access that requires manual approval and activation, removing the need to run the assume command twice.

### How did you test it?
See below.

### Potential risks
This changes the retry logic to use a constant retry rather than Fibonacci. There is a possibility that the constant attempt at requesting access could cause 429 errors, but in testing this did not happen (over a 10 minute period)

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -429,11 +429,12 @@ func AssumeCommand(c *cli.Context) error {
 			}
 
 			noAccessInput := accessrequesthook.NoAccessInput{
-				Profile:  profile,
-				Reason:   reason,
-				Duration: apiDuration,
-				Confirm:  assumeFlags.Bool("confirm"),
-				Wait:     wait,
+				Profile:   profile,
+				Reason:    reason,
+				Duration:  apiDuration,
+				Confirm:   assumeFlags.Bool("confirm"),
+				Wait:      wait,
+				StartTime: time.Now(),
 			}
 			retry, hookErr := hook.NoAccess(c.Context, noAccessInput)
 			if hookErr != nil {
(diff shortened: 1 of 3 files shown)


chrnorm (changes requested):
@meyerjrr this is looking good, I am thinking though the UX will be better if instead of

```
[i] Waiting for request to be approved and activated...
```

Granted instead did

```
[i] Waiting for request to be approved...
```

And then the Granted CLI should automatically activate it once it's approved.

meyerjrr:
UX now waits for approval and activates automatically when approved
Screenshot 2024-06-13 at 11 59 44 AM

If the user cannot self activate, this is the UX 
Screenshot 2024-06-13 at 11 59 57 AM

chrnorm (changes requested):
Looks good and the added log messages look good. I added an elapsed time printout to the logs.

I am consistently getting the following error message though, `Request was approved but failed to activate, user might not have permission to activate. Waiting for activation`:

```
❯ dassume mockaccount-2H12sU5WMHcIoKxPMXQmHhgRH5L/S3ListBuckets --wait --reason "test --wait"
[i] You don't currently have access to mockaccount-2H12sU5WMHcIoKxPMXQmHhgRH5L/S3ListBuckets, checking if we can request access...	[target=AWS::Account::"987564764670", role=S3ListBuckets, url=https://internal.commonfate.io]
[WILL REQUEST] S3ListBuckets access to mockaccount-2H12sU5WMHcIoKxPMXQmHhgRH5L will require approval
[REQUESTED] S3ListBuckets access to mockaccount-2H12sU5WMHcIoKxPMXQmHhgRH5L requires approval: https://internal.commonfate.io/access/requests/req_2hpCeRBVcEc8iaetirPq4irvnfb
[i] Waiting for request to be approved... [0s elapsed]
[i] Waiting for request to be approved... [10s elapsed]
[i] Request was approved but failed to activate, user might not have permission to activate. Waiting for activation. [30s elapsed]
[✔] [mockaccount-2H12sU5WMHcIoKxPMXQmHhgRH5L/S3ListBuckets](ap-southeast-2) session credentials will expire in 1 hour
```

There may be a race condition or a logic issue here, could you take a look into this before we merge?

Additionally, adding `GRANTED_LOG=debug` didn't give any verbose info, it may be helpful in debugging to add debug-level logs of the API response so that we can get a better understanding of what is happening

chrnorm:
I've also tweaked the error message to be slightly more user friendly, so it now reads `Request was approved but failed to activate, you might not have permission to activate. You can try and activate the access using the Common Fate web console.`

That said, my issue from above is still present

Issue #380, opened 2023-03-08
Granted gives an error when used with Firefox Developer Edition


chrnorm:
> When running assume -c I’m now getting this message:
Granted was unable to open a browser session automatically due to the following error: starting process: fork/exec {homedir}/Applications/Firefox Developer Edition.app: permission denied
with a URL to open directly in my browser which does work
I’ve got Firefox Developer Edition set as my browser, done by pasting in the path to the executable for granted browser set. I’ve checked the permissions on the executable, and it is rwxr-xr-x so it should be executable by all (edited)

This may be due due to the fork/exec and open changes made in https://github.com/common-fate/granted/pull/375

perpil:
Related: #366

PR #685, opened 2024-06-12
Add support for Firefox Nightly

meuxx:
### What changed?
Added support for Firefox Nightly

### Why?
Couldn't get granted to work with Firefox Nightly
Selecting Firefox and manually setting Firefox Nightly's path would produce the same error as in #380

### How did you test it?
1. `make cli`
2. `dgranted browser set` and select the new option Firefox Nightly
3. `dassume -c`
4. Granted opens Firefox Nightly with the right session on its own container

Only tested on MacOS but I checked that the binary paths for other OS are correct

### Potential risks
None

Yes, no breaking changes

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -464,7 +464,7 @@ func AssumeCommand(c *cli.Context) error {
 			return err
 		}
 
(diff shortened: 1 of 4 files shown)


chrnorm (approved):
Thankyou @meuxx! 🙌

PR #690, opened 2024-06-21
Handle nested parameters in granted settings set command

meyerjrr:
### What changed?
The `granted settings set` command will now handle nested parameters. For example `Backend.Keychain`. Dot notation is used for setting a nested value.

eg. `granted settings set --setting Keyring.Backend --value File`

Nested attributes will also show up in the list when running `granted settings set` without any flags:
 
```
❯ granted settings set
? Select the configuration to change  [Use arrows to move, type to filter]
> DefaultBrowser
  CustomBrowserPath
  CustomSSOBrowserPath
  Keyring.Backend
  Ordering
  ExportCredentialSuffix
  ExportCredsToAWS
```
### Why?
Previously there was no way to set these nested programatically and the `~/.granted/config`
 file would need to be manually edited.
### How did you test it?
`dgranted settings set` and setting the value of Keyring.Backend to a new value

`dgranted settings set --setting=Keyring.Backend --value=File`

```
❯ dgranted settings set --setting=Keyring.Backend --value=File
[i] Updating the value of Keyring.Backend to File
[✔] Config updated successfully
```
### Link to relevant docs PRs

### DIFF
--- a/pkg/granted/settings/set.go
+++ b/pkg/granted/settings/set.go
@@ -24,50 +24,7 @@ var SetConfigCommand = cli.Command{
 			return err
 		}
 
-		// Get the type and value of the Config struct
-		configType := reflect.TypeOf(*cfg)
-		configValue := reflect.ValueOf(cfg).Elem()
-		type field struct {
-			ftype  reflect.StructField
-			fvalue reflect.Value
-		}
-		var fields []string
-		var fieldMap = make(map[string]field)
-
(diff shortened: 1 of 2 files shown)


JoshuaWilkes:
In testing I found that the settings feature doesn't allow you to configure settings that don't already have a value.
Specificly if the go type in a pointer, it won;t show up in the list. See this screenshot of the Keychain field not being available because it is not set in the current config

image

I added a test case to show this not working

chrnorm (approved):

PR #691, opened 2024-07-02
Granted Doctor Command

meyerjrr:
### What changed?
Adds diagnostic tooling which will allow users to better understand the issues with granted / their local configuration with AWS 

the granted doctor command checks the existence of plaintext AWS SSO cached tokens and check the keychain for cached tokens, then run aws sts get-caller-identity on each cached profile to check if it's valid.

I have also added some frequent offenders of confusion that users run into, mostly surrounding Granteds configuration options.

### Why?
To help Granted users debug issues they may be having requesting access with Granted

### How did you test it?
run `granted doctor` in different scenarios with cached 

### Link to relevant docs PRs

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,7 +25,7 @@ require (
 	github.com/common-fate/clio v1.2.3
 	github.com/common-fate/common-fate v0.15.13
 	github.com/common-fate/glide-cli v0.6.0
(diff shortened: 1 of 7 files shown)


chrnorm (approved):
Confirmed working when tested locally

PR #694, opened 2024-07-03
Fix default duration when requesting access with Granted

chrnorm:
### What changed?
Fixes the default duration when requesting access with Granted - extracted this change from #691 so it can be merged separately, as it depends on an SDK change which needs to be approved.

### Why?
Granted uses the maximum duration for JIT access rather than the default duration

### How did you test it?
Untested locally

### Potential risks
May affect JIT access durations

Yes


meyerjrr (approved):
Approving but will require the next release of CF to go out before including this in a new Granted release.

PR #695, opened 2024-07-10
added flag to disable cache lookup and saving

meyerjrr:
### What changed?
Adds `--no-cache` flag which disables saving and pulling cached tokens from storage and forces a relogin. 

### Why?
Some users have found stale sso tokens are being pulled blocking them from getting valid credentials using sso. This is a potential work around for when this occurs

### How did you test it?
Using assume with the --no-cache flag and having to login.
Omitting the flag assume works as normal

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/entrypoint.go
+++ b/pkg/assume/entrypoint.go
@@ -54,7 +54,7 @@ func GlobalFlags() []cli.Flag {
 		&cli.StringFlag{Name: "reason", Usage: "Provide a reason for requesting access to the role"},
 		&cli.BoolFlag{Name: "confirm", Aliases: []string{"y"}, Usage: "Skip confirmation prompts for access requests"},
 		&cli.BoolFlag{Name: "wait", Usage: "When using Granted with Common Fate the assume will halt while waiting for the access request to be approved."},
-		&cli.BoolFlag{Name: "no-cache", Usage: "Disables caching of session credentials and forces a refresh"},
+		&cli.BoolFlag{Name: "no-cache", Usage: "Disables caching of session credentials and forces a refresh", EnvVars: []string{"GRANTED_NO_CACHE"}},
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/assume/entrypoint.go
```diff
@@ -54,6 +54,7 @@ func GlobalFlags() []cli.Flag {
 		&cli.StringFlag{Name: "reason", Usage: "Provide a reason for requesting access to the role"},
 		&cli.BoolFlag{Name: "confirm", Aliases: []string{"y"}, Usage: "Skip confirmation prompts for access requests"},
 		&cli.BoolFlag{Name: "wait", Usage: "When using Granted with Common Fate the assume will halt while waiting for the access request to be approved."},
+		&cli.BoolFlag{Name: "no-cache", Usage: "Disables caching of session credentials and forces a refresh"},
```
chrnorm, line 57: We should allow an env var to be used here too

```suggestion
		&cli.BoolFlag{Name: "no-cache", Usage: "Disables caching of session credentials and forces a refresh", EnvVars: []string{"GRANTED_NO_CACHE"}},
```

chrnorm (changes requested):
Could you please also ensure this works for the `granted credential-process` command?

i.e. calling `GRANTED_NO_CACHE=true granted credential-process <profile-name>` twice should return two sets of separate credentials.

Currently it returns a cached session.

meyerjrr:
Updated to add the same flag to the granted credential process binary. Note that there is also `DisableCredentialProccessCache` which is a config level setting for this same functionality. Which means there could be undesired / confusing ux if a user has one set and not the other.

pkg/granted/credential_process.go
```diff
@@ -55,7 +56,8 @@ var CredentialProcess = cli.Command{
 		secureSessionCredentialStorage := securestorage.NewSecureSessionCredentialStorage()
 		clio.Debugw("running credential process with config", "profile", profileName, "url", c.String("url"), "window", c.Duration("window"), "disableCredentialProcessCache", cfg.DisableCredentialProcessCache)
 
-		useCache := !cfg.DisableCredentialProcessCache
+		cliNoCache := c.Bool("no-cache")
+		useCache := !cfg.DisableCredentialProcessCache || !cliNoCache
```
JoshuaWilkes, line 60: I think you need to fix this logic so that useCache is false if either the config is set or the flag is set.
```suggestion
		useCache := !(cfg.DisableCredentialProcessCache || cliNoCache)
```

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

Issue #688, opened 2024-06-17
Config Filed does not exist


jrwpatterson:
Hi I am trying to use your wonderful software and I've used it tonnes in the past. 

But a weird thing is happening. Its logging in, but when it comes back, I'm getting an error of 

```
[✘] config file does not exist
```

No information why thats happening I've searched your repo and I can't find that anywhere in the code base! Its killing me!

The same config works on other users machines. I've deleted, reinstalled, nuked the cache, nuked the session token 

To make matters weirder it works with some configs that are identical bar the roll name or account number

meyerjrr:
Hi @jrwpatterson! Would you be able to run what is causing your issues with the GRANTED_LOG env variable set to debug? This will help in diagnosing where the config file read is getting called.

eg.
```
export GRANTED_LOG=debug && assume 
```

jrwpatterson:
Ok I'll have a look Monday seems to happen sometimes and sometimes it's fine!

douglaspetrin:
this is probably happening because you're missing the .aws/credentials file, or perhaps in your config file you are trying to download it but somehow it is not able to. for i.e: under your profile in your .aws/config file you have something like this to download your credentials:

[profile my_profile]
credential_process = granted credential-process --auto-login --profile my_profile

see: https://docs.commonfate.io/granted/recipes/credential-process

jrwpatterson:
sorry stuck again 

```[i] To assume this profile again later without needing to select it, run this command:
> assume Lasoo-AdministratorAccess-prod --env
2024/07/18 15:15:32 [keyring] Considering backends: [keychain]
2024/07/18 15:15:32 [keyring] Querying keychain for service="granted-aws-sso-tokens", account="https://d-.awsapps.com/start/base", keychain="login.keychain"
2024/07/18 15:15:32 [keyring] No results found
[DEBUG] error retrieving IAM Identity Center token from secure storage: The specified item could not be found in the keyring
2024/07/18 15:15:32 [keyring] Considering backends: [keychain]
2024/07/18 15:15:32 [keyring] Checking keychain status
2024/07/18 15:15:32 [keyring] Keychain status returned nil, keychain exists
2024/07/18 15:15:32 [keyring] Keychain item trusts keyring
2024/07/18 15:15:32 [keyring] Adding service="granted-aws-sso-tokens", label="", account="https://d-.awsapps.com/start/base", trusted=true to osx keychain "login.keychain"
[...]

PR #700, opened 2024-07-17
The NoAccess hook should not return an error when a user does not use Common Fate

JoshuaWilkes:
### What changed?
Brings the assume command in line with the credential process behaviour when a user does not have Common Fate deployment configured. Will now log the config error to debug and skip the check for an active request.

### Why?
Users reported that they were getting a config file error which is related to Common Fate despite them not having A deployment configured.

Our code was incorrectly handling an error loading a config file for CF in the assume command, despite it working for the credential process.

### How did you test it?
existing behaviour
```
assume Audit/AWSAdministratorAccess
[✘] config file does not exist
```

new behaviour
```
❯ dassume Audit/AWSAdministratorAccess
[✘] no access: operation error SSO: GetRoleCredentials, https response error StatusCode: 403, RequestID: bed86fbe-ec59-407d-8e50-399acea0acb2, api error ForbiddenException: No access

```

### Link to relevant docs PRs

### DIFF
--- a/pkg/hook/accessrequesthook/accessrequesthook.go
+++ b/pkg/hook/accessrequesthook/accessrequesthook.go
@@ -41,7 +41,8 @@ type NoAccessInput struct {
 func (h Hook) NoAccess(ctx context.Context, input NoAccessInput) (retry bool, err error) {
 	cfg, err := cfcfg.Load(ctx, input.Profile)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


JoshuaWilkes:
Fixes #688

chrnorm (approved):

PR #703, opened 2024-07-18
fix issue causing credential process to break

meyerjrr:
### What changed?
Bugfix: Fixed issue when duplicate profiles are found and the duplicate prefix was not applied to the credential process when using Granted registry 

### Link to relevant docs PRs

### DIFF
--- a/pkg/granted/awsmerge/merge_from_registry.go
+++ b/pkg/granted/awsmerge/merge_from_registry.go
@@ -189,6 +189,12 @@ func (m Merger) WithRegistry(src *ini.File, dst *ini.File, opts RegistryOpts) (*
 			return nil, err
 		}
 
+		//after the section has been copied, update the credential process if it exists to the new profile name
+
+		if f.HasKey("credential_process") {
+			f.Key("credential_process").SetValue(strings.Replace(f.Key("credential_process").Value(), sec.Name(), f.Name(), 1))
+		}
+
 		if f.Comment == "" {
 			f.Comment = "# profile name has been prefixed due to duplication"
 		} else {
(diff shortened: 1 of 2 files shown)


pkg/granted/registry/cfregistry/cfregistry.go
```diff
@@ -37,11 +37,16 @@ func (r *Registry) getClient() (awsv1alpha1connect.ProfileRegistryServiceClient,
 	if r.client != nil {
 		return r.client, nil
 	}
-	cfg, err := config.LoadDefault(context.Background())
+
+	err := config.SwitchContext(r.opts.Name)
```
chrnorm, line 41: This needs to be removed from this PR and put into a separate PR, it will break the existing behaviour because users have a `default` key for the CF file context, it doesn't match the profile registry

chrnorm (changes requested):

pkg/granted/registry/cfregistry/cfregistry.go
```diff
@@ -37,11 +37,16 @@ func (r *Registry) getClient() (awsv1alpha1connect.ProfileRegistryServiceClient,
...
 	if err != nil {
 		return nil, err
 	}
```
chrnorm, line 44:
```suggestion
```

pkg/granted/awsmerge/merge_from_registry.go
```diff
@@ -189,6 +189,12 @@ func (m Merger) WithRegistry(src *ini.File, dst *ini.File, opts RegistryOpts) (*
 			return nil, err
 		}
 
+		//after the section has been copied, update the credential process if it exists to the new profile name
```
chrnorm, line 192:
```suggestion
		// after the section has been copied, update the credential process if it exists to the new profile name
```

chrnorm (approved):

PR #708, opened 2024-07-22
Output actual credentials file path when exporting

dnrce:
### What changed?
Update log output from `assume --export` to show the actual path of the credentials file rather than the static value `~/.aws/credentials`.

### Why?
The credentials file path is not always `~/.aws/credentials`. It can be overridden with the environment variable `AWS_SHARED_CREDENTIALS_FILE`.

### How did you test it?
```
make cli
dassume --export
```

### Potential risks
No functional risks identified.

User experience risk that users prefer a more compact path containing a tilde instead of a full home directory.

Yes, does not change API.

### Link to relevant docs PRs
N/A

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -578,8 +578,8 @@ func AssumeCommand(c *cli.Context) error {
 				clio.Warn("No credential suffix found. This can cause issues with using exported credentials if conflicting profiles exist. Run `granted settings export-suffix set` to set one.")
 			}
 
-			clio.Successf("Exported credentials to ~/.aws/credentials file as %s successfully", profileName)
-
+			credentialsFilePath := cfaws.GetAWSCredentialsPath()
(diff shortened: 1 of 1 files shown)


pkg/assume/assume.go
```diff
@@ -578,8 +578,8 @@ func AssumeCommand(c *cli.Context) error {
 				clio.Warn("No credential suffix found. This can cause issues with using exported credentials if conflicting profiles exist. Run `granted settings export-suffix set` to set one.")
 			}
 
-			clio.Successf("Exported credentials to ~/.aws/credentials file as %s successfully", profileName)
-
+			credentialsFilePath := cfaws.GetAWSCredentialsPath()
```
dnrce, line 581: This made for the smallest PR, but if preferred I could offer a version in which `cfaws.ExportCredsToProfile()` returns the path so this is less implementation-aware.

chrnorm:
Thanks so much @dnrce! 🙌

chrnorm (approved):

PR #719, opened 2024-07-29
feat: support Flatpak Firefox

p5:
### What changed?

Initial support for Flatpak-based browsers, in this PR, this is scoped to only Firefox.

This implementation treats Flatpak as a browser, which is invoked with the app ID.  It doesn't make too much sense logically, but it follows the existing conventions.  Otherwise it would have required far larger changes I am unable to commit to at this time.

### Why?

Linux users don't always have a browser installed as a system package.  It's getting more and more common to use Flatpaks installed from Flathub (you can think of it as the Linux app store)

### How did you test it?

On a machine with a Flatpak Firefox installation, I ran `make cli`, `dgranted set browser` and selected "Flatpak - Firefox".  
This prompts the user to enter the app ID (since it can therefore support all variations of Flatpak Firefox (dev, nightly etc).

Once that is done, I validated the `~/.dgranted/config` file contained the required parameters:
```
DefaultBrowser = "FLATPAK_FIREFOX"
CustomBrowserPath = "/usr/bin/flatpak"
CustomSSOBrowserPath = ""
FlatpakBrowserAppID = "org.mozilla.firefox"
FlatpakSSOBrowserAppID = ""
Ordering = ""
ExportCredentialSuffix = ""
```

And finally, I ran `gdassume -c <role>` which opened up a Flatpak Firefox browser.

### Link to relevant docs PRs

### DIFF
--- a/pkg/browser/detect.go
+++ b/pkg/browser/detect.go
@@ -50,7 +50,21 @@ func HandleManualBrowserSelection() (string, error) {
 	withStdio := survey.WithStdio(os.Stdin, os.Stderr, os.Stderr)
 	in := survey.Select{
 		Message: "Select one of the browsers from the list",
-		Options: []string{"Chrome", "Brave", "Edge", "Firefox", "Waterfox", "Chromium", "Safari", "Stdout", "FirefoxStdout", "Firefox Developer Edition", "Firefox Nightly", "Arc"},
+		Options: []string{
+			"Chrome",
+			"Brave",
+			"Edge",
+			"Firefox",
+			"Waterfox",
+			"Chromium",
+			"Safari",
+			"Stdout",
+			"FirefoxStdout",
+			"Firefox Developer Edition",
+			"Firefox Nightly",
+			"Arc",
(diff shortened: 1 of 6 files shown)


pkg/browser/detect.go
```diff
@@ -100,40 +114,46 @@ func Find() (string, error) {
...
 		return FirefoxDevEditionKey
 	}
 
-	if strings.ToLower(b) == "firefox nightly" {
+	if strings.EqualFold(b, FirefoxNightlyKey) {
 		return FirefoxNightlyKey
 	}
-
-	if strings.Contains(strings.ToLower(b), "brave") {
+	if strings.HasPrefix(strings.ToLower(b), "flatpak") {
```
p5, line 128: This was placed higher up the list since it was conflicting with the "mozilla" check.  

Looking again, I think this function needs a small refactor, since keys are upper cased and separated with underscores, whereas `b` is regular space-separated text.

p5:
Superceeded by #731

PR #722, opened 2024-08-05
Access Request should not immediately extend after request has just been activated

ckluy31:
### What changed?
Previously, access requests that had just been activated would automatically be extended, since the batch ensure API is called twice. This was not an issue before we deprecated `try_extend_after_seconds`, which would not allow a user to extend until x seconds after activation. However, since that has now been deprecated, it lead to this undesirable behaviour.

This PR will not attempt another batch ensure api call if the grant has just been activated.

### How did you test it?

Testing evidence:

Before it would immediately extend an activated request when using assuming.

image

After the changes, when running the assume command:

```
calvinluy➜~/Git/granted(calvin/cf-3453-access-extension-does-not-work✗)» dassume                                                          [16:41:21]

? Please select the profile you would like to assume: CommonFateDemo/AWSReadOnlyAccess                
[i] To assume this profile again later without needing to select it, run this command:
> assume CommonFateDemo/AWSReadOnlyAccess 
[i] You don't currently have access to CommonFateDemo/AWSReadOnlyAccess, checking if we can request access...   [target=AWS::Account::"975050289140", role=AWSReadOnlyAccess, url=https://internal.commonfate.io]
[WILL ACTIVATE] AWSReadOnlyAccess access to CommonFateDemo will be activated for 1h: https://internal.commonfate.io/access/requests/req_2kEEiG50qjxxt3selokLIOzKtj5
[i] Access::Grant::"gra_2kEEiEFtWkRPQE5y4aC8AbL7M3E": read only AWS access is automatically approved
? Reason for access (Required) Test that it should not automatically extend the request upon activation.
[i] Access::Grant::"gra_2kEElPHPwGD1tp3e0dS06BM64lJ": read only AWS access is automatically approved
[ACTIVATED] AWSReadOnlyAccess access to CommonFateDemo was activated for 1h: https://internal.commonfate.io/access/requests/req_2kEElJqAxlVCOmAii8ZLhFZM2EU
[i] Access::Grant::"gra_2kEElPHPwGD1tp3e0dS06BM64lJ": read only AWS access is automatically approved
[✔] [CommonFateDemo/AWSReadOnlyAccess](ap-southeast-2) session credentials will expire in 1 hour
```

It will behave as expected:
Screenshot 2024-08-05 at 4 49 06 PM

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -357,7 +357,7 @@ func AssumeCommand(c *cli.Context) error {
 				Wait:      wait,
 				StartTime: time.Now(),
 			}
-			retry, hookErr := hook.NoAccess(c.Context, noAccessInput)
+			retry, justActivated, hookErr := hook.NoAccess(c.Context, noAccessInput)
 			if hookErr != nil {
 				return hookErr
 			}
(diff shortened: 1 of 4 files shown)


JoshuaWilkes (approved):
Looks good.

I reckon we will still have an edge case, pretty minor where you activate the request first then use assume but before the AWS credentials are ready.
probably a window of ~5 seconds. It would auto extend the access.

I think we shoudl be looking at reintroducing a fix in the Common Fate API for this too

PR #732, opened 2024-08-13
Open the common fate console

JoshuaWilkes:
### What changed?
Adds `granted cf` which can be used to launch the Common Fate console from the cli.
It first tries to detect a console url from the aws config, if more than one is found, a select is shown.
If none is found in the profiles, falls back to the cf config file.

a `--profile` flag can be used to open the console for5 a specific profile

### Why?
convenient way to get to the console to make a request using the web when needed

### How did you test it?
Tested with more then one console URL in my config

### Potential risks
This does not respect any custom browser configuration

### Link to relevant docs PRs

### DIFF
--- a/pkg/granted/cf.go
+++ b/pkg/granted/cf.go
@@ -1,0 +1,54 @@
+package granted
+
+import (
+	"github.com/common-fate/clio"
+	"github.com/common-fate/granted/pkg/cfaws"
+	"github.com/common-fate/granted/pkg/cfcfg"
+	sdkconfig "github.com/common-fate/sdk/config"
+	"github.com/pkg/browser"
+	"github.com/urfave/cli/v2"
+)
+
+var CFCommand = cli.Command{
+	Name:  "cf",
+	Usage: "Open Common Fate in the browser",
+	Action: func(c *cli.Context) error {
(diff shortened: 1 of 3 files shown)


pkg/granted/cf.go
```diff
@@ -0,0 +1,89 @@
...
+				return err
+			}
+			url, err := cfcfg.GetCommonFateURL(p)
+			if err != nil {
+				return err
+			}
+			if url == nil {
+				return errors.New("the profile exists but it is not configured with with a Common Fate console url")
+			}
+			consoleURL = url.String()
```
meyerjrr, line 42: If `--profile` flag is provided, what is the UX here? It looks like it will continue down and still ask which deployment to open. Should this be returning early here?

```
consoleURL = url.String()
return browser.OpenURL(consoleURL)
```

JoshuaWilkes: Fixed this now

meyerjrr (approved):

chrnorm (changes requested):
Thinking `cf` may not be self-explanatory that it is going to open a browser window, what about `granted web open` instead?

PR #748, opened 2024-09-17
Fix using default duration on auto approved requests

meyerjrr:
### What changed?
Updated the access request hooks to use the Duration returned by the batch ensure API to show how long access was requested for. Where a request is already active, the remaining duration is calculated.

### Why?
Access requests that required access to be assigned with Common Fate was returning incorrect values for the duration on the request.

### How did you test it?
Ran assume without the change and observed the bug present.
Reran the same command and got the expected output of the default duration. 

With specifying duration
```
dassume  dev.test/Billing -d 1h
[i] You don't currently have access to dev.test/Billing, checking if we can request access... [target=AWS::Account::"123456789012", role=Billing, url=http://localhost:9090]
[WILL ACTIVATE] Billing access to test will be activated for 1h: http://localhost:8080/access/requests/req_2mBcrs8p9mF9Y7ZqaCEQcaoF7ZR
? Reason for access (Required) test
[ACTIVATED] Billing access to test was activated for 1h: http://localhost:8080/access/requests/req_2mBcsCc6S5C0Z0dPyHMdXRQRBYz
[✔] [dev.test/Billing](ap-southeast-2) session credentials will expire in 1 hour
```

When not specifying duration, defaults to using the default duration
```
dassume  dev.test/Billing 
[i] You don't currently have access to dev.test/Billing, checking if we can request access... [target=AWS::Account::"123456789012", role=Billing, url=http://localhost:9090]
[WILL ACTIVATE] Billing access to test will be activated for 5m: http://localhost:8080/access/requests/req_2mBcw1qVFNVrRfxVZ13nc8iuhj6
? Reason for access (Required) test
[ACTIVATED] Billing access to test was activated for 5m: http://localhost:8080/access/requests/req_2mBcwLqKRDxTUaEKq53Lt8FoQj8
```

### Link to relevant docs PRs

### DIFF
--- a/pkg/hook/accessrequesthook/accessrequesthook.go
+++ b/pkg/hook/accessrequesthook/accessrequesthook.go
@@ -206,19 +206,8 @@ func (h Hook) NoEntitlementAccess(ctx context.Context, cfg *config.Context, inpu
 	for _, g := range res.Msg.Grants {
 		names[eid.New("Access::Grant", g.Grant.Id)] = g.Grant.Name
 
-		exp := "<invalid expiry>"
-
-		if req.Entitlements[0].Duration != nil {
-			exp = ShortDur(req.Entitlements[0].Duration.AsDuration())
-		} else if res.Msg.DurationConfiguration != nil {
-			exp = ShortDur(res.Msg.DurationConfiguration.MaxDuration.AsDuration())
-			if res.Msg.DurationConfiguration.DefaultDuration != nil {
-				exp = ShortDur(res.Msg.DurationConfiguration.DefaultDuration.AsDuration())
-			}
-		} else if g.Grant.ExpiresAt != nil {
-			//attempt to work out duration from expiry to preserve backwards compatability with older common fate versions
-			exp = ShortDur(time.Until(g.Grant.ExpiresAt.AsTime()))
-		}
+		// default is to show the original duration, except for an active request, where it gets recalculated below to the time remaining
+		exp := ShortDur(g.Grant.Duration.AsDuration())
 
 		switch g.Change {
 		case accessv1alpha1.GrantChange_GRANT_CHANGE_ACTIVATED:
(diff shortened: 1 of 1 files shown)


pkg/hook/accessrequesthook/accessrequesthook.go
```diff
@@ -166,6 +166,19 @@ func (h Hook) NoEntitlementAccess(ctx context.Context, cfg *config.Context, inpu
 			req.Justification.Reason = &customReason
 		}
 	}
+
+	for i := range req.Entitlements {
+
+		if input.Duration != nil {
+			req.Entitlements[i].Duration = input.Duration
+
+		} else if result.DurationConfiguration.DefaultDuration != nil {
```
JoshuaWilkes, line 175: Looks like there needs to be another check here for DurationConfiguration being set to avoid a nil pointer
`if res.Msg.DurationConfiguration != nil`

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

JoshuaWilkes:
I've updated this PR to use the Duration field which was added a few months ago, this now reports exactly how long the grant was requested for, or how long is remaining on an active grant.

pkg/hook/accessrequesthook/accessrequesthook.go
```diff
@@ -166,6 +166,24 @@ func (h Hook) NoEntitlementAccess(ctx context.Context, cfg *config.Context, inpu
...
+
+		if result.DurationConfiguration != nil {
+			if result.DurationConfiguration.DefaultDuration != nil {
+				req.Entitlements[i].Duration = result.DurationConfiguration.DefaultDuration
+			} else {
+				req.Entitlements[i].Duration = result.DurationConfiguration.MaxDuration
+			}
+		}
+
+	}
```
JoshuaWilkes, line 185:
```suggestion
	for i := range req.Entitlements {

		if input.Duration != nil {
			req.Entitlements[i].Duration = input.Duration
			continue
		}

		if result.DurationConfiguration != nil {
			if result.DurationConfiguration.DefaultDuration != nil {
				req.Entitlements[i].Duration = result.DurationConfiguration.DefaultDuration
			} else {
				req.Entitlements[i].Duration = result.DurationConfiguration.MaxDuration
			}
		}

	}
```

Only set the duration on the grant when it has been provided as a flag, else, the server will work out the duration.
```suggestion
        if input.Duration != nil {
	        for i := range req.Entitlements {
			req.Entitlements[i].Duration = input.Duration
		}
	}
```

JoshuaWilkes (changes requested):

pkg/hook/accessrequesthook/accessrequesthook.go
JoshuaWilkes, line 185: The duration flag already gets set on the request at the top of the file.
If it is not provided the server chooses the correct duration.
```suggestion
```

JoshuaWilkes (changes requested):

JoshuaWilkes (approved):

PR #751, opened 2024-09-19
update file permissions to read/write

meyerjrr:
### What changed?
Updated file permissions from 700 to 644

### Why?
Some of the files and folders that Granted created for config were using a file perm of 700 which is for read/write/execute. This is unnecessary for their use.

### Link to relevant docs PRs

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -17,6 +17,16 @@ import (
 	"github.com/common-fate/granted/internal/build"
 )
 
+const (
+	// permission for user to read/write.
+	USER_READ_WRITE_PERM = 0644
+)
+
+const (
+	// permission for user to read/write.
+	USER_READ_WRITE_EXECUTE_PERM = 0700
+)
+
 type BrowserLaunchTemplate struct {
 	// UseForkProcess specifies whether to use forkprocess to launch the browser.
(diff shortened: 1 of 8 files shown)


pkg/config/config.go
```diff
@@ -137,7 +137,7 @@ func SetupConfigFolder() error {
 		return err
 	}
 	if _, err := os.Stat(grantedFolder); os.IsNotExist(err) {
-		err := os.Mkdir(grantedFolder, 0700)
+		err := os.Mkdir(grantedFolder, 0644)
```
JoshuaWilkes, line 140: @meyerjrr for all of these, I think it s would be good to add them as a const in the file where they are used with the comment on read/Write

for any files that have the constant but it is unused, remove it

```suggestion
		err := os.Mkdir(grantedFolder, USER_READ_WRITE_PERM)
```

JoshuaWilkes (changes requested):

pkg/config/config.go
JoshuaWilkes: If we have other file perms in use, you coud follow the same pattern of adding them as consts

JoshuaWilkes (approved):

Issue #754, opened 2024-09-24
Config file perrmisions appear to be wrong


jpbelleau:
It appears the file permissions for the Granted config directory, when created, are wrong.  We saw the change in behavior with the 0.34 release.  If fails when running `granted registry add ...`.  Looking at #751, the wrong permissions were used for directory creation.

chrnorm:
Thanks for flagging this @jpbelleau, I'm working on a fix for this now

jpbelleau:
I almost have I have one done and working to open a PR.

chrnorm:
Thankyou. My proposed change is to simply revert #751

jpbelleau:
I have not sent a change to granted yet and have some setup to do.  I found the 2 offending issues and no revert should be needed.

jpbelleau:
Added PR for the config folder fixes

chrnorm:
@jpbelleau this is now released in [v0.34.1](https://github.com/common-fate/granted/releases/tag/v0.34.1), thanks again for the report. The release binaries are now building and will be live in the next 10 minutes or so. I ended up merging my PR over yours because I wanted to shift the permissions back to `0600`. I marked you as a coauthor on the PR though to highlight your contribution - thankyou so much for your help here.

Some context too on why #751 was implemented at all - @JoshuaWilkes noticed recently that that the `~/.granted/config` file was executable on his machine, and the PR was put up as a response to this. Upon reviewing the change though I think that the Josh's `~/.granted/config` file may have been made executable outside of Granted, as I was unable to reproduce this myself.

PR #755, opened 2024-09-24
Revert "update file permissions to read/write (#751)"

chrnorm:
### What changed?
Reverts the change made in #751.

### Why?
#751 incorrectly removed the executable flag on some of the directories created by Granted. The execute flag [is required](https://superuser.com/questions/168578/why-must-a-folder-be-executable). Additionally, the PR relaxed some of the permissions from `0600` to `0644` which allows users other than the current one to read the various config and frecency files. Given that we've never had a permissions issue opened due to a use case where `0644` is required, I think this is too permissive.

### How did you test it?

Build the CLI locally:

```
sudo make cli
```

Move the `~/.dgranted` folder:

```
mv ~/.dgranted ~/.dgranted.bak
```

Run `dassume`, which will run through the onboarding wizard.

Confirm that the `config` file is not executable and has the expected permissions:

```
❯ ls -al ~/.dgranted
drwx------ chrisnorman staff 128 B  Tue Sep 24 17:35:51 2024  .
drwxr-x--- chrisnorman staff 3.5 KB Tue Sep 24 17:35:51 2024  ..
.rw------- chrisnorman staff 200 B  Tue Sep 24 17:35:54 2024  config
.rw------- chrisnorman staff 1.7 KB Tue Sep 24 17:35:55 2024  log
```

### Potential risks

Low, as we are reverting a known issue in the new release. 

Yes

### DIFF
--- a/pkg/accessrequest/role.go
+++ b/pkg/accessrequest/role.go
@@ -14,11 +14,6 @@ import (
 	"github.com/common-fate/granted/pkg/config"
 )
 
-const (
-	// permission for user to read/write.
-	USER_READ_WRITE_PERM = 0644
-)
-
 type Role struct {
 	Account string `json:"account"`
 	Role    string `json:"role"`
@@ -51,7 +46,7 @@ func (r Role) Save() error {
 	}
 
 	file := filepath.Join(configFolder, "latest-role")
-	return os.WriteFile(file, roleBytes, USER_READ_WRITE_PERM)
+	return os.WriteFile(file, roleBytes, 0644)
 }
 
 func LatestRole() (*Role, error) {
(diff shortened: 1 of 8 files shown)


chrnorm:
Fixes #754. A big thankyou to [jpbelleau](https://github.com/jpbelleau) for helping to diagnose this issue so quickly.

chrnorm:
I also removed most of the constants like `USER_READ_WRITE_PERM` here - using SCREAMING_SNAKE_CASE makes these constants publicly available for each package which isn't really desirable. Additionally, I found that in a few cases the constants were only used in a single location, so it's easier IMO to just use the bitmask directly rather than adding a layer of indirection.

VigneshSelvaraj96 (approved):
Tested Locally and file permissions match as expected. Great catch on the screaming snake case.

PR #756, opened 2024-09-24
Fix config folder permissions on create

### LINKED ISSUE #754: Config file perrmisions appear to be wrong

jpbelleau:
Fixes #754

### What changed?
Updated config directory permissions to allow execute

### Why?
Allow directory to function

### How did you test it?
Verified locally

### Link to relevant docs PRs

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -23,7 +23,7 @@ const (
 )
 
 const (
-	// permission for user to read/write.
+	// permission for user to read/write/execute.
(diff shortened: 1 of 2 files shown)


chrnorm:
I ended up merging #755 instead and have marked you as a coauthor on that change - thankyou!

PR #757, opened 2024-09-24
Add RAM (Resource Access Manager)

wayne-folkes:
### What changed?
I added `ram` Resource Access Manager to the service map

### Why?
It is a service that exists and will get rid of this message ```[!] We don't recognize service ram but we'll try and open it anyway (you may receive a 404 page)```

### Link to relevant docs PRs


chrnorm (approved):
Thankyou @wayne-folkes! 🙌

Issue #739, opened 2024-08-23
Mangled AWS config output when certain characters are present in values


Lerdson:
Trying to introduce Granted to my workplace but I've run in to an AWS config writing bug that's currently stopping us from storing our AWS configs in a Granted profile registry.

## Background
We have a number of entries using SAML auth following this format:
```
[profile foo:dev]
azure_app_id_uri             = https://signin.aws.amazon.com/saml\#0000000000042
```

When Granted parses and writes these entries out in operations such as registry syncing, the parser picks up the "#" character as a comment and decides to convert it from an inline comment in to a standalone line comment like so:
```
[profile foo:dev]
# 0000000000042
azure_app_id_uri             = https://signin.aws.amazon.com/saml
```
As you would expect, this breaks the authentication process as the endpoint has been mangled. Quoting the value does not appear to help.

I was able to trace this issue in to Granted's use of https://github.com/go-ini/ini, which currently does not respect escaped symbols such as "\#", in contrast with AWS CLI's own parser. I believe this will also affect configs containing "\;" looking at the code. In my experiments I attempted the naive fix of rebuilding Granted to pass the`IgnoreInlineComment`. When passed to calls of `ini.Empty()` and `ini.LoadSources()` this appears to fix the issue at the cost of losing inline comment compatibility.

There does not appear to be another way without patching the ini library code to have smarter handling of escaped special characters. This would perhaps be best solved on the ini library side but nonetheless I thought I'd log the issue here as it does directly affect Granted.

### Repository adding/syncing
1. Create a Granted profile registry with a config with values containing "\#".
2. Add/Sync from the registry using `granted registry add`.
3. Observe the generated sections in your local AWS config file.

### Creating a new profile registry from your local AWS config.
1. Create a Granted profile registry with a config with values containing "\#".
2. Initialise a new profile registry from your existing AWS config using `granted registry setup`.
3. Observe the generated registry config file.

PR #766, opened 2024-10-02
Handle parsing of special comment characters in configuration

ckluy31:
### What changed?
In order to fix issue https://github.com/common-fate/granted/issues/739, the ini client is loaded with the following options:

- SpaceBeforeInlineComment : Only interprets the inline comment if there is a space preceding the # character
- UnescapeValueCommentSymbols : Prevents the # from being automatically escaped (i.e. \#) as part of the output

### Why?
Users were unable to use URLs that included the # character, which was typically used for inline comments. A thorough description of the issue can be seen here https://github.com/common-fate/granted/issues/739.

Fixes https://github.com/common-fate/granted/issues/739.

### How did you test it?

Using the dummy configuration with # characters in git registry https://github.com/common-fate/granted-registry/blob/main/config2:
```
# this is a dummy config 
[profile demo-main_dummy]
granted_sso_start_url = https://d-976708da7d.awsapps.com/start\#0000000000042
granted_sso_account_id = 632700053629
granted_sso_role_name = AWSAdministratorAccess
region = ap-southeast-2
session_name = {{ .Required.SessionName }}
credential_process = granted credential-process --profile {{ .Profile }} --url https://internal.prod.granted.run
```

We then generate the registries in `.aws/config` using `dgranted registry add -name test12 -url https://github.com/common-fate/granted-registry/ -prefix-duplicate-profiles`

The output observed in  `.aws/config` is:

```
...

[profile test12.demo-main_dummy]
granted_sso_start_url  = `https://d-976708da7d.awsapps.com/start#0000000000042`
granted_sso_region     = ap-southeast-2
granted_sso_account_id = 632700053629
granted_sso_role_name  = AWSAdministratorAccess
region                 = ap-southeast-2
session_name           =
credential_process     = granted credential-process --profile test12.demo-main_dummy --url https://internal.prod.granted.run

[granted_registry_end test12]
```

Furthermore, when creating a new profile registry from your local AWS config, this is the generated granted.yml:

```
[profile test12.demo-main_dummy]
granted_sso_start_url  = `https://d-976708da7d.awsapps.com/start#0000000000042`
granted_sso_region     = ap-southeast-2
granted_sso_account_id = 632700053629
granted_sso_role_name  = AWSAdministratorAccess
region                 = ap-southeast-2
session_name           =
credential_process     = granted credential-process --profile test12.demo-main_dummy --url https://internal.prod.granted.run
```

### Potential risks
Previous configurations that do not have a space before the inline comment will not have it recognised as an inline comment. Instead, it will be recognised as part of the value.

### Link to relevant docs PRs

### DIFF
--- a/pkg/granted/registry/cfregistry/cfregistry.go
+++ b/pkg/granted/registry/cfregistry/cfregistry.go
@@ -91,7 +91,12 @@ func (r *Registry) AWSProfiles(ctx context.Context) (*ini.File, error) {
 		}
 	}
 
-	result := ini.Empty()
+	result := ini.Empty(
+		ini.LoadOptions{
+			SpaceBeforeInlineComment:    true,
+			UnescapeValueCommentSymbols: true,
+		},
+	)
 
 	for _, profile := range profiles {
 
--- a/pkg/granted/registry/gitregistry/gitregistry.go
+++ b/pkg/granted/registry/gitregistry/gitregistry.go
@@ -58,7 +58,12 @@ func (r Registry) AWSProfiles(ctx context.Context) (*ini.File, error) {
 
 	// load all cloned configs of a single repo into one ini object.
 	// this will overwrite if there are duplicate profiles with same name.
-	result := ini.Empty()
+	result := ini.Empty(
+		ini.LoadOptions{
+			SpaceBeforeInlineComment:    true,
+			UnescapeValueCommentSymbols: true,
+		},
+	)
 
 	for _, cfile := range cfg.AwsConfigPaths {
 		var filepath string
--- a/pkg/granted/sso.go
+++ b/pkg/granted/sso.go
@@ -44,7 +44,7 @@ var SSOCommand = cli.Command{
 
 const (
 	// permission for user to read/write.
-	USER_READ_WRITE_PERM = 0644 
+	USER_READ_WRITE_PERM = 0644
 )
 
 // in dev:


chrnorm:
I tested this locally, I think this might cause us problems because backticks are not supported by the AWS CLI's config parser.

For example, the following configuration is invalid:

```
[profile example-backtick]
sso_start_url  = `https://d-12345abcdef.awsapps.com/start`
sso_region     = ap-southeast-2
sso_account_id = 123456789012
sso_role_name  = AWSAdministratorAccess
```

With the following error:
```
❯ assume

? Please select the profile you would like to assume: example-backtick
[!] error retrieving IAM Identity Center token from secure storage: The specified item could not be found in the keyring
[✘] operation error SSO OIDC: StartDeviceAuthorization, https response error StatusCode: 400, RequestID: 6bd61cdf-4aa7-419c-8fdd-786a96fd12dc, InvalidRequestException:
```

Essentially this is caused by a difference between the current ini package we use, and the Python [ConfigParser](https://docs.python.org/3/library/configparser.html#supported-ini-file-structure) package used in the AWS CLI. We might want to look at https://github.com/bigkevmcd/go-configparser as an alternative library

chrnorm:
I realised we actually swapped __away__ from the configparser library in this PR: https://github.com/common-fate/granted/pull/297

PR #770, opened 2024-10-07
Improve how RDS proxy command handles grant expiry and proxy connection errors

JoshuaWilkes:
### What changed?
This PR resolves some issues with handling the case where the proxy server terminates a connection due to any number of reasons, like grant expiry or an inaccessible target database.

The CLI client now listens for a signal that the session has closed and prints a friendly error message.

Implementation wise, I've done a major refactor to split the command up into smaller functions to improve readability.
I've also removed some go routines which were flakey and instead the code now waits for specific conditions like the SSM connection starting up and being ready without.

I've improved the handling of the client connection listener so it now switches on either the session being closed or a new connection coming in.

### Why?
Confusing error messages and the proxy would hang sometimes due to the SSM handler not shutting down as expected.

### How did you test it?
Locally against a production database.

### Link to relevant docs PRs


chrnorm (approved):
confirmed working for me in a local build

PR #775, opened 2024-10-09
Changes waf shortcut to use v2 home page, rather than WAF classic home.

CodyDunlap:
### What changed?

Added wafv2 service map

### Why?

Because the WAF v2 home page is actually at wafv2/homev2

### Potential risks

None

Yes


chrnorm (approved):
Thankyou @CodyDunlap!

PR #792, opened 2024-11-05
Make leading/trailing spaces optional in templating

agershman:
### What changed?

Allow templating in granted config file to contain lead/trailing spaces or not at all.

### Why?

The usage of templating in granted config files is more forgiving if the author doesn't have spaces in their go templating syntax. Without this it's unclear why template syntax is left un-rendered in the AWS config file.

### How did you test it?

Built CLI and tested using a registry containing a test profile config and granted config file:

#### Test profile config

image

#### Test granted config

image

#### Add test registry

image

#### Observe rendered profiles

image

### Potential risks

There shouldn't be risks given that the updated regular expression allows for zero or more spaces, which should be backwards compatible with all existing usages.

### Link to relevant docs PRs

N/A

### DIFF
--- a/pkg/granted/awsmerge/merge_from_registry.go
+++ b/pkg/granted/awsmerge/merge_from_registry.go
@@ -303,7 +303,7 @@ func interpolateVariables(value string, profileName string) (string, error) {
 }
 
 func containsTemplate(text string) bool {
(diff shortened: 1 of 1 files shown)


chrnorm (approved):
Thanks @agershman!

PR #797, opened 2024-11-08
Add Vivaldi browser support

jpts:
### What changed?

Added support for Vivaldi browser

### Why?

Allows users to more easily select the browser.

### How did you test it?

1. `go build -ldflags='-w -s -X github.com/common-fate/granted/internal/build.ConfigFolderName=.granted' ./cmd/granted`
2. `granted browser set`
3. `assume <profile> -c`

Tested on MacOS and Linux. Windows path was inferred from [forum posts](https://forum.vivaldi.net/search?term=vivaldi.exe&in=titlesposts).

### Potential risks

None

Yes, no breaking changes

### Link to relevant docs PRs

### DIFF
--- a/scripts/assume
+++ b/scripts/assume
@@ -11,7 +11,7 @@ _this_type=$(type -- "${0##*/}" 2>&1)
 # In the case of zsh, the output will contain the word 'alias'.
 # shellcheck disable=SC3028
 if [ "${_this_type#*not found}" != "$_this_type" ] ||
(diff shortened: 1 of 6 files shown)


scripts/assume
JoshuaWilkes, line 1: These probably weren't intended to be included but I'll keep it in because these whitespaces look like they shouldn't be there.

JoshuaWilkes (approved):

Issue #785, opened 2024-10-22
Failure to save aws config after registry sync run


meyerjrr:
A user has reported not being able to run and complete a registry sync run. This actually caused them not to be able to request access due to the auto sync functionality.

```Error when retrieving credentials from custom-process: [i] Syncing Profile Registries
panic: runtime error: index out of range [0] with length 0

goroutine 1 [running]:
gopkg.in/ini%2ev1.(*File).writeToBuffer(0x140005c0700, {0x0, 0x0})
        /Users/runner/go/pkg/mod/gopkg.in/ini.v1@v1.67.0/file.go:352 +0xe40
gopkg.in/ini%2ev1.(*File).SaveToIndent(0x140005c0500?, {0x14000133278, 0x17}, {0x0?, 0x0?})
        /Users/runner/go/pkg/mod/gopkg.in/ini.v1@v1.67.0/file.go:530 +0x30
gopkg.in/ini%2ev1.(*File).SaveTo(...)
        /Users/runner/go/pkg/mod/gopkg.in/ini.v1@v1.67.0/file.go:540
github.com/common-fate/granted/pkg/granted/registry.SyncProfileRegistries({0x104259dc0, 0x140007184b0}, 0x0)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/granted/registry/sync.go:112 +0x574
github.com/common-fate/granted/pkg/autosync.runSync({0x104259dc0, 0x140007184b0}, {{0x1400004a800?, 0x103112f00?}, 0x10408dd40?}, 0x0)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/autosync/registry.go:22 +0x6c
github.com/common-fate/granted/pkg/autosync.Run({0x104259dc0, 0x140007184b0}, 0x0)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/autosync/autosync.go:44 +0x120
github.com/common-fate/granted/pkg/granted.GetCliApp.WithAutosync.func4(0x1?)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/granted/middleware/sync.go:10 +0x2c
github.com/common-fate/granted/pkg/granted.GetCliApp.WithBeforeFuncs.func5(0x1400063c240)
        /Users/runner/work/granted-cli-build/granted-cli-build/granted/pkg/granted/middleware/cli.go:26 +0x84
[...]

PR #805, opened 2024-11-15
Clean up registry package

EloyTolosaDev:
It started because I wanted to investigate on issue #785 , and I ended up refactoring the codebase a bit while working on it. In the end, I did nothing about the issue because I couldn't figure out what was happening, and I did not get any response from the person who created the issue, so I could not investigate further. After all, it looks like a user problem, as I was unable to reproduce the problem.

### What changed?
Functionality did not change, the main changes were:
1. Removing  an unused function "IsOutdatedConfig" as it was only checking if the length of the variable ProfileRegistryURLs and that could be done separately without any function.
2. Removed an unused error variable "RegistrySyncError" as it did not provide any extra functionality in the case of the error ocurred, also it was not being used.
3. Removed the awsmerge.Merger struct, and replacing the merger struct functions as package functions, as there were no differences and having an extra awsmerge.Merger struct which did nothing at all looked strange to me.

### How did you test it?
I tested the dassume and dgranted binaries and worked as espected.

### Link to relevant docs PRs


pkg/autosync/autosync.go
```diff
@@ -12,20 +12,19 @@ import (
...
-		return
-	}
 
 	registries, err := registry.GetProfileRegistries(interactive)
 	if err != nil {
 		clio.Debugf("unable to load granted config file with err %s", err.Error())
 		return
 	}
 
+	if len(registries) > 0 {
```
chrnorm, line 22: `registry.IsOutdatedConfig()` checks a different field in the Granted config to the one in `registry.GetProfileRegistries()`.

IsOutdatedConfig checks if the length of `ProfileRegistryURLS` is > 0, whereas `GetProfileRegistries` uses the `ProfileRegistry` field.

The change you have made here will cause valid profile registry configuration to return the error. Could you please restore the previous behaviour? 

In this instance, I'd propose just restoring the `registry.IsOutdatedConfig` function and restoring the call to the function. The code there was working in production, so there is no need to change it IMO.

chrnorm (changes requested):
Thanks for this contribution @EloyTolosaDev! One requested change here

EloyTolosaDev:
Hi! Thanks for the response. I checked what you said and you're totally true, I was confused about profileRegistryURLs field and the actual loadedRegistry structs.
I changed and checked that so now everything should be ok.

Kind regards

chrnorm (approved):
Thankyou @EloyTolosaDev!

PR #810, opened 2024-12-02
Add gimme-aws-creds assumer

jpts:
### What changed?

Add support for the [gimme-aws-creds](https://github.com/Nike-Inc/gimme-aws-creds) helper as an assumer, which supports AWS accounts integrated with Okta directly via SAML. The program supports multiple versions of Okta, MFA configurations etc. so we don't have to reimplement any of that logic.

On top of the basic support, it has the following features:
- caches obtained temporary credentials via the secure storage mechanism
- parses the gimme-aws-creds config file, so we can use other credential providers for non-matching profiles
- limited support for headless credential_process based refresh. It depends on an opt in from the user (using either the --auto-login flag or config file setting), and a compatible version of the tool and Okta flow. It's opt-in because the user can't verify the device code, since it's hidden by the calling program.

### Why?

Allows us to get credentials via Okta/SAML, which wasn't previously supported.

### How did you test it?

granted compiled with `go build -ldflags='-w -s -X github.com/common-fate/granted/internal/build.ConfigFolderName=.granted' ./cmd/granted`

Tested over the last few months on multiple AWS accounts/orgs, including a mixed config with Identity Centre and other credential providers.

### Potential risks

None

Yes, no breaking changes

### Link to relevant docs PRs

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -69,6 +69,7 @@ require (
 	github.com/google/uuid v1.6.0 // indirect
 	github.com/gorilla/securecookie v1.1.2 // indirect
 	github.com/gorilla/websocket v1.5.3 // indirect
+	github.com/hashicorp/go-version v1.7.0 // indirect
 	github.com/huandu/xstrings v1.3.3 // indirect
 	github.com/imdario/mergo v0.3.11 // indirect
 	github.com/invopop/yaml v0.2.0 // indirect
--- a/pkg/cfaws/assumer_aws_gimme_aws_creds.go
+++ b/pkg/cfaws/assumer_aws_gimme_aws_creds.go
@@ -21,6 +21,8 @@ import (
 const GRANTED_OKTA_INI_OPEN_BROWSER = "granted_okta_open_browser"
 
 type AwsGimmeAwsCredsAssumer struct {
+	config           *ini.File
+	forceOpenBrowser bool
 }
 
 type CredentialCapture struct {
@@ -81,9 +83,16 @@ func (gimme *AwsGimmeAwsCredsAssumer) AssumeTerminal(ctx context.Context, c *Pro
 		return *creds, nil
 	}
 
-	// if cred process fail
+	// if cred process, check we can do a non-interactive refresh
 	if configOpts.UsingCredentialProcess {
-		return aws.Credentials{}, fmt.Errorf("Cannot refresh Gimme AWS creds in credential_process")
+		err := gimme.LoadGimmeConfig()
+		if err != nil {
+			return aws.Credentials{}, fmt.Errorf("Failed to load gimme config file: %w", err)
+		}
+		if !gimme.CanRefreshHeadless(c.Name) {
+			return aws.Credentials{}, fmt.Errorf("Cannot refresh Gimme AWS creds in credential_process when force_classic is set.")
+		}
+		gimme.forceOpenBrowser = true
 	}
 
 	clio.Debugw("refreshing credentials", "reason", "none cached")
@@ -101,8 +110,12 @@ func (gimme *AwsGimmeAwsCredsAssumer) AssumeTerminal(ctx context.Context, c *Pro
 		}
 
 		if ob.MustBool(false) == true || ob.String() == "true" {
-			args = append(args, "--open-browser")
+			gimme.forceOpenBrowser = true
 		}
+	}
+
+	if gimme.forceOpenBrowser {
+		args = append(args, "--open-browser")
 	}
 
 	// add passthrough args
@@ -165,8 +178,29 @@ func (gimme *AwsGimmeAwsCredsAssumer) Type() string {
 	return "AWS_GIMME_AWS_CREDS"
 }
 
-// inspect for any items on the profile prefixed with "granted_okta"
+// parse the gimme config file to check if we have a matching profile
 func (gimme *AwsGimmeAwsCredsAssumer) ProfileMatchesType(rawProfile *ini.Section, parsedProfile config.SharedConfig) bool {
+	err := gimme.LoadGimmeConfig()
+	if errors.Is(err, os.ErrNotExist) {
+		return false
+	}
+	if err != nil {
+		clio.Error("Failed to load gimme config file: ", err)
+		return false
+	}
+
+	for _, section := range gimme.config.SectionStrings() {
+		if section == parsedProfile.Profile {
+			clio.Debug("matched gimme profile ", section)
+			return true
+		}
+	}
+
+	//clio.Debug("No gimme profile matched")
+	return false
+}
+
+func (gimme *AwsGimmeAwsCredsAssumer) LoadGimmeConfig() error {
 	okta_config := os.Getenv("OKTA_CONFIG")
 	if okta_config == "" {
 		home, err := os.UserHomeDir()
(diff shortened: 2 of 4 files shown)


chrnorm (commented):
Hey @jpts, thanks so much for the contribution! I've taken a look over it and have some initial feedback.

For context: long term, I'd actually like to remove the `Assumer` interface entirely. I believe that some of the third-party assumers are only used by one or two Granted users, and it seems like this interface adds complexity to Granted and additional code to maintain.

I'm thinking eventually that we'll replace this with some sort of hook - e.g. allow you to call a bash script which could then trigger `gimme-aws-creds` (or any other sort of login tooling).

In the meantime though I'd be happy to accept additional assumer contributions until this change is implemented. Your PR here is good because it doesn't affect any of the core session management code etc.

It might actually be better here if we could add a `bash_assumer` that would trigger a bash script - would this work to solve your use case? I'm just thinking that this might be a better step towards my long term goal of using hooks and eliminating the custom `Assumer` code altogether. Let me know if this would work for your use case though.

Separately I have a few comments specific to the code - I'll add these inline on your diff!

go.mod
```diff
@@ -32,6 +32,7 @@ require (
 	github.com/common-fate/sdk v1.71.0
 	github.com/common-fate/xid v1.0.0
 	github.com/fatih/color v1.16.0
+	github.com/hashicorp/go-version v1.7.0
```
chrnorm, line 35: I'd like to minimise the number of third party dependencies Granted uses, is it possible for you to vendor the code you need from `go-version` or implement this with a simple regex rather than using a third-party library?

(this comment is hypocritical because the go.mod currently has quite a few third-party deps 🤣 but I would like to thin these out over time!)

pkg/cfaws/assumer_aws_gimme_aws_creds.go
```diff
@@ -0,0 +1,293 @@
...
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/common-fate/clio"
+	"github.com/common-fate/granted/pkg/securestorage"
+	"github.com/hashicorp/go-version"
+	"gopkg.in/ini.v1"
+)
+
+const GRANTED_OKTA_INI_OPEN_BROWSER = "granted_okta_open_browser"
```
chrnorm, line 22: I'd prefer to avoid adding additional `granted_` config flags for assumers - I'm thinking here for the initial implementation we could just choose a default rather than allowing this to be customisable? i.e. have it always open the browser perhaps? Let me know if this would work for your use case - I'm not sure if you've been using this internally and have had a need for both behaviours.

My main concern is that it's difficult to track these config flags if they are spread out across separate assumers.

pkg/cfaws/assumer_aws_gimme_aws_creds.go
```diff
@@ -0,0 +1,293 @@
...
+		if err != nil {
+			return aws.Credentials{}, fmt.Errorf("could not parse credentials expiry: %s", cc.result.Credentials.Expiration)
+		}
+		c.Expires = t
+	}
+	return c, nil
+}
+
+func (gimme *AwsGimmeAwsCredsAssumer) AssumeTerminal(ctx context.Context, c *Profile, configOpts ConfigOpts) (aws.Credentials, error) {
+	// try cache
```
chrnorm, line 76: ~~The caching logic is handled elsewhere in Granted already, this cache checking code _shouldn't_ be needed (we should treat it as a separate bug in Granted if this is not the case)~~

Ok upon looking at the other assumers, they all implement the caching logic themselves. This is actually a reason I'd like to remove this interface as the session caching should be shared by all assumers. Just adding this as a note so you have some context on my thinking here, but no changes required here.

pkg/cfaws/assumer_aws_gimme_aws_creds.go
```diff
@@ -0,0 +1,293 @@
...
+	capture := &CredentialCapture{}
+	cmd.Stdout = capture
+	cmd.Stdin = os.Stdin
+	cmd.Stderr = os.Stderr
+
+	cleanEnv := []string{}
+	var disallowedVar bool
+	for _, env := range os.Environ() {
+		disallowedVar = false
+		for _, disallowed := range []string{
```
chrnorm, line 139: Why do you strip these env variables? (might be good to add a comment here to explain the implementation)

pkg/cfaws/assumer_aws_gimme_aws_creds.go
```diff
@@ -0,0 +1,293 @@
...
+func (gimme *AwsGimmeAwsCredsAssumer) AssumeConsole(ctx context.Context, c *Profile, configOpts ConfigOpts) (aws.Credentials, error) {
+	return gimme.AssumeTerminal(ctx, c, configOpts)
+}
+
+func (gimme *AwsGimmeAwsCredsAssumer) Type() string {
+	return "AWS_GIMME_AWS_CREDS"
+}
+
+// parse the gimme config file to check if we have a matching profile
+func (gimme *AwsGimmeAwsCredsAssumer) ProfileMatchesType(rawProfile *ini.Section, parsedProfile config.SharedConfig) bool {
```
chrnorm, line 186: is it possible to read a key on the AWS profile itself rather than parsing the gimme config file here? AFAIK, this code will be executed for all Granted users for every profile so Granted is going to try to read the gimme config file on everyone's machine every time `assume` is called

pkg/cfaws/assumer_aws_gimme_aws_creds.go
```diff
@@ -0,0 +1,293 @@
...
+
+	gimme.config, err = ini.Load(okta_config)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (gimme *AwsGimmeAwsCredsAssumer) CanRefreshHeadless(profile string) bool {
```
chrnorm, line 243: Could we support only 2.6.0+ here? It looks like multi-version support makes this more complex, do you have users in your team internally spread across multiple versions?

chrnorm (changes requested):
Added a few questions for you on the implementation!

go.mod
jpts: Happy to switch to regex if you'd prefer. I understand the desire to keep the 3rd party dependencies to a minimum. I think the tradeoff is that a simple regex won't support the full semver syntax, so it's a bit more brittle.

pkg/cfaws/assumer_aws_gimme_aws_creds.go
jpts: That's fine by me, it's not something we've been heavily using.

pkg/cfaws/assumer_aws_gimme_aws_creds.go
jpts: Ack. I did peruse the other assumers to see if I could re-use any of the logic, but nothing seemed suitable. Happy to adapt if anything changes.

pkg/cfaws/assumer_aws_gimme_aws_creds.go
jpts: It's definitely possible, however I think it would require a new `granted_` key on the AWS profile. If I understood correctly, this means I would have to manually go and add this key to every AWS profile I wanted to use with this assumer? It seemed more elegant to me to check if the gimme config file exists, which is what the current implementation does.

NB. We check if the file exists first. If it doesn't, it doesn't try to load/parse it.

pkg/cfaws/assumer_aws_gimme_aws_creds.go
jpts: It's designed to make `gimme-aws-creds` execute more reliably. We never want it to use existing credentials, always to fetch new ones. I'll add a comment.

pkg/cfaws/assumer_aws_gimme_aws_creds.go
jpts: I think we have very few users on these early versions, so I'd personally be happy to drop support for them. My only concern is how we would communicate that support to users. ie. if we removed this check, I think the headless refresh would hang or just crash and return a non-descriptive error.

jpts:
@chrnorm thanks for the review and context around the direction of the tool. I've left my replies inline.

A new `bash_assumer` could work if it exposed the same sort of interfaces as the current assumer code. I don't think this assumer is particularly unique or special, so a generic hook/integration should work well.

go.mod
chrnorm: Thanks. After some more thought let's keep this third-party dep, as you note the regex will be more brittle.

pkg/cfaws/assumer_aws_gimme_aws_creds.go
chrnorm: Thanks - I'm happy with your approach here so let's keep it as-is.

pkg/cfaws/assumer_aws_gimme_aws_creds.go
chrnorm: Let's leave this as-is!

chrnorm (commented):
Thanks @jpts, just the comment on the additional config flag is outstanding and once that is resolved this will be ready to merge!

chrnorm (approved):

chrnorm:
Thanks @jpts!

set92:
I'm missing some documentation, or tutorial on how to use this, right?

Issue #807, opened 2024-11-21
Add support for opening multiple tabs using `--services` cli option


cannonkalra:
While opening browser console, open tabs for multiple services

`assume -c staging -s ec2 -s ecs -s iam -s s3`

liam-baker-tpc:
This would be very helpful.
Alternative syntax could be `assume -c staging -s ec2,ecs,iam,s3`

PR #827, opened 2025-02-07
Added ability to specify multiple services top be opened with the --console flag

meyerjrr:
### What changed?
Altered the services flag to now be a string slice flag which accepts multiple services (ie `assume -c -s s3 -s rds`) Which will then open up the console with two tabs one for each of the services.

To do this I altered the logic for creating URLs to now create a list of URL's even when just creating one.

It then loops through all urls created and runs the same browser logic previously used.

### Why?
I've been learning vim and have been looking for simple tasks to complete on Granted :D
For #807

### How did you test it?
Manually testing `assume -c` with and without the -s flag for one and multiple services

Running related commands like `assume --url` `assume` `granted console` to confirm no regressions made

### Potential risks
string -> stringSlice flag should have no usage changes but all code paths using .String() need to have been changed to ensure correct logic

Altering logic in the assume command has the potential to create regressions in some of the flags logic.

It might be worth noting here that the `assume.go` file is becoming quite complicated due to small additions over time and could use with a refactor to simplify it down. This would be a big task but could be useful to minimise any regressions being added.

### Link to relevant docs PRs

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -415,7 +415,7 @@ func AssumeCommand(c *cli.Context) error {
 			return err
 		}
 
-		clio.Debugf(`number of console urls created`, "amount", len(consoleURLs))
+		clio.Debugf("number of console urls created", "amount", len(consoleURLs))
 		for _, consoleURL := range consoleURLs {
 
 			if cfg.DefaultBrowser == browser.FirefoxKey || cfg.DefaultBrowser == browser.WaterfoxKey || cfg.DefaultBrowser == browser.FirefoxStdoutKey || cfg.DefaultBrowser == browser.FirefoxDevEditionKey || cfg.DefaultBrowser == browser.FirefoxNightlyKey {
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -330,10 +330,14 @@ func AssumeCommand(c *cli.Context) error {
 	if getConsoleURL {
 		con := console.AWS{
 			Profile:     profile.Name,
-			Service:     assumeFlags.String("service"),
-			Region:      region,
 			Destination: assumeFlags.String("console-destination"),
 		}
+
+		services := assumeFlags.StringSlice("service")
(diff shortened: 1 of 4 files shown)


pkg/assume/assume.go
```diff
@@ -406,92 +410,98 @@ func AssumeCommand(c *cli.Context) error {
...
+		consoleURLs, err := con.URLs(creds)
 		if err != nil {
 			return err
 		}
 
-		if cfg.DefaultBrowser == browser.FirefoxKey || cfg.DefaultBrowser == browser.WaterfoxKey || cfg.DefaultBrowser == browser.FirefoxStdoutKey || cfg.DefaultBrowser == browser.FirefoxDevEditionKey || cfg.DefaultBrowser == browser.FirefoxNightlyKey {
-			// transform the URL into the Firefox Tab Container format.
-			consoleURL = fmt.Sprintf("ext+granted-containers:name=%s&url=%s&color=%s&icon=%s", containerProfile, url.QueryEscape(consoleURL), profile.CustomGrantedProperty("color"), profile.CustomGrantedProperty("icon"))
-		}
+		clio.Debugf(`number of console urls created`, "amount", len(consoleURLs))
```
chrnorm, line 418: format nit
```suggestion
		clio.Debugf("number of console urls created", "amount", len(consoleURLs))
```

pkg/console/aws.go
```diff
@@ -28,11 +28,39 @@ type awsSession struct {
 	SessionToken string `json:"sessionToken"`
 }
 
+func (a AWS) URLs(creds aws.Credentials) ([]string, error) {
+
+	urls := []string{}
+
+	//if region and service were not specified create a single default url and return
+	if a.Region == "" && a.Service == nil {
```
chrnorm, line 36: this check will be safer
```suggestion
	if a.Region == "" && len(a.Service) == 0 {
```

chrnorm (changes requested):
thankyou @meyerjrr this is great! a couple of minor changes requested

PR #877, opened 2025-05-16
Adding cloudfront, elasticache and sqs to service list.

felipewnp:


felipewnp:
@chrnorm can we do something to pass these changes?

PR #908, opened 2025-11-30
fix: credential process ignores --mfa-token flag

### LINKED ISSUE #604: --mfa-token not registering from CLI, only from prompt
The `--mfa-token` switch does not seem to work when using a `source_profile`. But it does work by inputting it on the prompt.

Below I use the TOTP seed to generate an OTP. I call `assume` with the `--mfa-token` switch, but `assume` still prompts for an OTP. The `--mfa-token` does not seem to register the OTP parameter. But the OTP works by putting it in the prompt.

```
❯ export OTP=$(echo 0000O73YQPSHRX6DNWOCD6| totp-cli instant) && echo $OTP
743051
❯ assume lxk-sandbox --mfa-token 743051
? MFA Token 743051
[✔] [lxk-sandbox](us-east-1) session credentials will expire in 1 hour
```

My profiles look like
```
[profile lxk-iam]
region             = us-east-1
credential_process = granted credential-process --profile=lxk-iam

[profile lxk-sandbox]
role_arn       = arn:aws:iam::00000000000:role/@Global_Administrator
source_profile = lxk-iam
region         = us-east-1
mfa_serial     = arn:aws:iam::111111111111:mfa/mfa-cli
```

dafujii:
Fixed #604

The `--mfa-token` CLI flag was being ignored when using `credential_process` with `source_profile` configurations, causing users to be prompted interactively for MFA tokens even when provided via the command line.

Updated `CredentialProcessAssumer.AssumeTerminal()` to check if `configOpts.MFATokenCode` is provided before falling back to `MfaTokenProvider`.

### DIFF
--- a/pkg/cfaws/assumer_aws_credential_process.go
+++ b/pkg/cfaws/assumer_aws_credential_process.go
@@ -49,10 +49,22 @@ func (cpa *CredentialProcessAssumer) AssumeTerminal(ctx context.Context, c *Prof
 				}
 				if p.AWSConfig.MFASerial != "" {
 					aro.SerialNumber = &p.AWSConfig.MFASerial
-					aro.TokenProvider = MfaTokenProvider
+					if configOpts.MFATokenCode != "" {
+						aro.TokenProvider = func() (string, error) {
+							return configOpts.MFATokenCode, nil
+						}
+					} else {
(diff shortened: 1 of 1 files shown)


chrnorm (approved):
Thankyou!

Issue #617, opened 2024-03-05
Unclear warning with no way to quiet the error text.


zwing99:
when you do not have a suffix you get this warning and there seems to be no way to silence the warning without setting a suffix. Furthermore the docs and warning itself do not make clear the risk. Please allow me to silence the warning AND please make it clear what i should be afraid of.

```bash
$ assume coolcompany/Developer
[✔] [coolcompany/Developer](us-east-1) session credentials will expire in 8 hours
[!] No credential suffix found. This can cause issues with using exported credentials if conflicting profiles exist. Run `granted settings export-suffix set` to set one.
[✔] Exported credentials to ~/.aws/credentials file as coolcompany/Developer successfully
```

https://github.com/common-fate/granted/blob/3ae786ea106e6ec82bb978a3e6f7a0fc381928c2/pkg/assume/assume.go#L531

chrnorm:
Thanks for the report @zwing99. Could you please share the `[profile coolcompany/Developer]` block in your `~/.aws/config` (redacting any AWS account IDs etc)?

zwing99:
```
[profile granappdevelopment/Developer]
granted_sso_start_url      = https://coolcompany.awsapps.com/start
granted_sso_region         = us-east-1
granted_sso_account_id     = 3.141592653589793 🤪
granted_sso_role_name      = Developer
common_fate_generated_from = aws-sso
credential_process         = granted credential-process --profile coolcompany/Developer
```

calebhailey:
I'm also curious about this. I'm working with a team that (so far) doesn't use or need suffixes. I have run `granted settings export-suffix set`, leaving the suffix blank, and that also doesn't remove the warning. I have also set the `GRANTED_QUIET=true` environment variable and this also doesn't remove the warning. Any suggestions?

PR #915, opened 2026-01-15
Allow ExportCredentialSuffix to be nil

supergibbs:
Allow ExportCredentialSuffix to be nil and only warns if so. This allows users to set a blank suffix to quiet the warning.

Also updates the warning from:

`No credential suffix found. This can cause issues with using exported credentials if conflicting profiles exist. Run 'granted settings export-suffix set'`

to:

`No credential suffix found. This can cause issues with using exported credentials if conflicting profiles exist. Run 'granted settings export-suffix set' to set one. Set to empty string to supress this warning`

I am not a go developer but I was able to test new feature with `dassume` and tests still pass. :)

Addresses #617

### DIFF
--- a/pkg/assume/assume.go
+++ b/pkg/assume/assume.go
@@ -464,12 +464,13 @@ func AssumeCommand(c *cli.Context) error {
 				return err
 			}
 			var profileName string
-			if cfg.ExportCredentialSuffix != "" {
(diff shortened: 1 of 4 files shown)


chrnorm (approved):
lgtm thankyou!

PR #567, opened 2023-12-01
Update mock browser to simulate issue #208

nocnokneo:
### What changed?

Update mock test browser to demonstrate issue #208 

### Link to relevant docs PRs

Fixes issue #208

PR #916, opened 2026-01-22
fix: silence SSO browser stdout

### LINKED ISSUE #208: On Ubuntu linux in Bash, credentials are only set after running `assume` twice
When running `assume my-profile` I get a success message
```
[cf-dev](ap-southeast-2) session credentials will expire 2022-08-04 11:50:47 +0800 AWST
```

However running `env | grep AWS` shows that no credentials are exported into my shell env

Running assume again for the same profile, and repeating the env check shows that credentials are exported.
Then aws commands work as expected

benjamin-pilgrim:
## Summary
- discard browser stdout/stderr so SSO browser noise doesn't pollute credential_process JSON
- include the mock browser stdout change from #567

Fixes #208

### DIFF
--- a/cmd/testing/browser/main.go
+++ b/cmd/testing/browser/main.go
@@ -1,10 +1,13 @@
 package main
 
 import (
+	"fmt"
 	"os"
 )
 
 // This package is used to test browser features in tests
 func main() {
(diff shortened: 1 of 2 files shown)


chrnorm (approved):
LGTM, thankyou

Issue #583, opened 2023-12-28
assume profile name autocompletion not working on fish shell


epoch-chrono:
```bash
fish --version
fish, version 3.6.4

granted --version

  /$$$$$$                                 /$$                     /$$
 /$$__  $$                               | $$                    | $$
| $$  \__/  /$$$$$$  /$$$$$$  /$$$$$$$  /$$$$$$    /$$$$$$   /$$$$$$$
| $$ /$$$$ /$$__  $$|____  $$| $$__  $$|_  $$_/   /$$__  $$ /$$__  $$
| $$|_  $$| $$  \__/ /$$$$$$$| $$  \ $$  | $$    | $$$$$$$$| $$  | $$
| $$  \ $$| $$      /$$__  $$| $$  | $$  | $$ /$$| $$_____/| $$  | $$
|  $$$$$$/| $$     |  $$$$$$$| $$  | $$  |  $$$$/|  $$$$$$$|  $$$$$$$
 \______/ |__/      \_______/|__/  |__/   \___/   \_______/ \_______/

Version: 0.20.5

assume --version

  /$$$$$$
 /$$__  $$
| $$  \ $$  /$$$$$$$ /$$$$$$$ /$$   /$$ /$$$$$$/$$$$   /$$$$$$
| $$$$$$$$ /$$_____//$$_____/| $$  | $$| $$_  $$_  $$ /$$__  $$
| $$__  $$|  $$$$$$|  $$$$$$ | $$  | $$| $$ \ $$ \ $$| $$$$$$$$
| $$  | $$ \____  $$\____  $$| $$  | $$| $$ | $$ | $$| $$_____/
| $$  | $$ /$$$$$$$//$$$$$$$/|  $$$$$$/| $$ | $$ | $$|  $$$$$$$
|__/  |__/|_______/|_______/  \______/ |__/ |__/ |__/ \_______/

Version: 0.20.5
```

![assume](https://github.com/common-fate/granted/assets/108760692/3d085957-2cff-42fe-a62e-0719228ff56b)

mironx:
Hi, I have similar problem on zsh.
I had to back to version v0.20.2.
With this older version v0.20.2 - assume is working with zsh.
Thank you.

PR #919, opened 2026-02-02
feat: add dynamic profile listing to fish shell autocomplete

allixsenos:
## Summary
- Fish shell completions previously used urfave/cli's static `ToFishCompletion()`, which only generated completions for subcommands and flags — AWS profile names were never listed
- Replaced with dynamic fish completion templates that call `assumego --generate-bash-completion` at tab-completion time, matching the existing zsh behavior
- Separate completion files are now generated for both `assume` and `granted` commands

## Test plan
- [x] Run `granted completion --shell fish` and verify files are created in `~/.config/fish/completions/`
- [x] In fish shell, type `assume ` + tab and verify AWS profiles from `~/.aws/config` and `~/.aws/credentials` are listed
- [x] Verify `assume -` + tab still lists flags
- [x] Verify `granted ` + tab still lists subcommands

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/granted/completion.go
+++ b/pkg/granted/completion.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/common-fate/clio"
 	"github.com/fwdcloudsec/granted/internal/build"
-	"github.com/fwdcloudsec/granted/pkg/assume"
 	"github.com/fwdcloudsec/granted/pkg/config"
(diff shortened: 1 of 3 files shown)


allixsenos:
fixes #583

JoshuaWilkes (approved):

PR #942, opened 2026-05-18
feat(assume): add --export-as option

### LINKED ISSUE #821: --export with a new name
Hi folks 👋 

Is there a way to export credentials but giving it a new name in the process?

Something like:

```
assume <profilname> --export-as <new-profilename>
```

Currently I have a profile that I use to request a token, and then use the resulting token to assume the other profiles. Currently I use the export, and change the name on the credential manually. I might be misunstanding a concept here completely, but this gets me to where I need to go.

If it is not something you currently support, would you be open to have me take a stab at adding it to granted? If I'm doing something wrong here, could someone please help me understand how to use granted for my scenario? 😄 

kind regards

QuentinBtd:
## Description
This PR implements the feature requested in #821 by adding support for exporting credentials to `~/.aws/credentials` under a custom profile name.

### What changed
- Added a new `assume` flag: `--export-as`
  - Example: `assume <source-profile> --export-as <new-profile-name>`
 
- Updated export flow to:
  - Trigger credential export when `--export-as` is provided (even without `--export`)
  - Use the provided custom profile name as-is
 
- Extended credential export implementation with:
  - `ExportCredsToProfileWithOptions(profileName, creds, applySuffix bool)`
 
### Behavior details
- Existing behavior is unchanged for `--export`:
  - `ExportCredentialSuffix` is still applied when configured.
- New behavior for `--export-as`:
  - `ExportCredentialSuffix` is intentionally **not** applied.
  - The destination profile name is exactly the value passed to `--export-as`.

Closes #821.

### DIFF
--- a/pkg/cfaws/cred_exporter.go
+++ b/pkg/cfaws/cred_exporter.go
@@ -13,6 +13,11 @@ import (
 
 // ExportCredsToProfile will write assumed credentials to ~/.aws/credentials with a specified profile name header
 func ExportCredsToProfile(profileName string, creds aws.Credentials) error {
+	return ExportCredsToProfileWithOptions(profileName, creds, true)
+}
+
+// ExportCredsToProfileWithOptions will write assumed credentials to ~/.aws/credentials with additional export options
+func ExportCredsToProfileWithOptions(profileName string, creds aws.Credentials, applySuffix bool) error {
 	// fetch the parsed cred file
 	credPath := GetAWSCredentialsPath()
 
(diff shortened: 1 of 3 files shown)


QuentinBtd:
Hello @meyerjrr, what do you think about this feature?

pkg/cfaws/cred_exporter.go
```diff
@@ -13,6 +13,11 @@ import (
 
 // ExportCredsToProfile will write assumed credentials to ~/.aws/credentials with a specified profile name header
 func ExportCredsToProfile(profileName string, creds aws.Credentials) error {
+	return ExportCredsToProfileWithOptions(profileName, creds, true)
+}
+
+// ExportCredsToProfileWithOptions will write assumed credentials to ~/.aws/credentials with additional export options
+func ExportCredsToProfileWithOptions(profileName string, creds aws.Credentials, applySuffix bool) error {
```
meyerjrr: We don't need to make a new function here. Can you just add `applySuffix bool` to the original `ExportCredsToProfile` function?

meyerjrr (commented):
logic make sense to me, one small code quality nit to fix up then I can merge this

pkg/cfaws/cred_exporter.go
QuentinBtd: Thanks for the feedback! 

It's done 👍

meyerjrr (approved):

PR #947, opened 2026-06-04
feat: add Helium browser support

byrafael:
Adds support for **Helium**, a Chromium-based browser for macOS. It follows the same integration pattern as Brave, Edge, and Vivaldi.

Changes

- `pkg/browser/browsers.go`: add `HeliumKey` constant, `HeliumPathMac` detection path, and `HeliumPathDefaults()`
- `pkg/browser/detect.go`: add Helium to the manual selection list, `GetBrowserKey` name mapping, and `DetectInstallation` switch
- `pkg/assume/assume.go`: route `HeliumKey` through the `ChromeProfile` launcher
- `pkg/granted/console.go`: route `HeliumKey` through the `ChromeProfile` launcher
- `pkg/launcher/chrome_profile.go`: add `HeliumPathMac` local state path (`Library/Application Support/net.imput.helium/Local State`) for profile switching support

Notes

**Helium** is macOS-only (no Windows or Linux paths), consistent with how Arc is handled. The local state path was confirmed against an actual Helium installation.

Testing

Verified on macOS with Helium installed at `/Applications/Helium.app` — granted browser set auto-detects the installation and profile switching works correctly.


JoshuaWilkes:
You will need to update pkg/launcher/launcher.go
Add the helium key there
```
func ForBrowser(key, path, profile string) (Launcher, error) {
	switch key {
	case browser.ChromeKey, browser.BraveKey, browser.EdgeKey, browser.ChromiumKey, browser.VivaldiKey, browser.HeliumKey:
```

PR #949, opened 2026-06-10
fix: silent SSO browser stdout when using default browser

kien-truong:
The fix in #916 forgot to redirect Stdout for default browser


pkg/idclogin/browser.go
```diff
@@ -32,6 +32,8 @@ func openBrowser(url string, browserProfile string) error {
 		return openWithCustomPath(config.CustomSSOBrowserPath, url, browserProfile)
 	}
 
+  defaultbrowser.Stdout = os.Stderr
```
JoshuaWilkes:
```suggestion
	defaultbrowser.Stdout = os.Stderr
```

JoshuaWilkes (approved):

PR #952, opened 2026-07-02
Add Linux Kernel keyring support

jpts:
Add Linux kernel keyring support. It's supported in the upstream keyring library, so we only have to enable it.

Useful for linux systems that don't support Gnome/KDE Secret Service.


jpts:
@sbldevnet @meyerjrr Any chance of a review on this one? 🙏

PR #965, opened 2026-08-02
fix(sso): honour dual-stack and FIPS endpoint settings

### LINKED ISSUE #960: Device code SSO login fails against dual-stack Identity Center instances, and `AWS_USE_DUALSTACK_ENDPOINT` is silently ignored
`assume` and `granted sso login` can't complete a device-code login against an Identity Center instance on the new dual-stack portal domain (`https://<instance>.portal.<region>.app.aws`). The link granted prints points at the legacy IPv4 portal on a different domain than the configured start URL, and sign-in there fails unless you already have a session on that domain.

## Repro

Profile pointing at a dual-stack instance:

```
[profile example]
granted_sso_start_url = https://ssoins-REDACTED.portal.us-east-2.app.aws
granted_sso_region    = us-east-2
granted_sso_account_id = 000000000000
granted_sso_role_name  = AdministratorAccess
```

`assume example` prints:

```
[i] If the browser does not open automatically, please open this link: https://ssoins-REDACTED.us-east-2.portal.amazonaws.com/#/device?user_code=XXXX-XXXX
```

Different domain than the start URL.

## Root cause

Two separate things.

granted is faithful — `pkg/idclogin/run.go` prints `deviceAuth.VerificationUriComplete` verbatim, and AWS is what returns the legacy domain. Confirmed with a raw unauthenticated call using the same start URL, where the only variable is which endpoint the request lands on:

```
$ aws sso-oidc start-device-authorization \
    --start-url https://ssoins-REDACTED.portal.us-east-2.app.aws --region us-east-2 ...
verificationUriComplete: https://ssoins-REDACTED.us-east-2.portal.amazonaws.com/#/device?user_code=RDKV-JGVM

$ AWS_USE_DUALSTACK_ENDPOINT=true aws sso-oidc start-device-authorization ...same args...
verificationUriComplete: https://ssoins-REDACTED.portal.us-east-2.app.aws/#/device?user_code=XVNK-HVKG
```

So AWS returns a usable URI if the sso-oidc call goes to the dual-stack endpoint. That half is tracked upstream at https://github.com/aws/aws-cli/issues/10507.

The granted-side bug is that there's no way to ask for it. `pkg/cfaws/assumer_aws_sso.go:193` and `pkg/granted/sso.go:311` build the client config as:

```go
newCfg := aws.NewConfig()
newCfg.Region = rootProfile.SSORegion()
```

`aws.NewConfig()` leaves `ConfigSources` empty, and the SDK bails out early:

```go
func resolveUseDualStackEndpoint(cfg aws.Config, o *Options) error {
	if len(cfg.ConfigSources) == 0 {
		return nil
	}
	...
}
```

(`service/ssooidc/api_client.go`, unchanged as of v1.35.15, the version pinned on main.)

So both `AWS_USE_DUALSTACK_ENDPOINT=true` and `use_dualstack_endpoint = true` in `~/.aws/config` are dropped before they reach the client. The documented aws-cli workaround has no effect on granted.

## Workaround

`UseAuthorizationCode = true` in `~/.granted/config`. PKCE never calls `StartDeviceAuthorization`, so the bad URI is never produced — it registers with `IssuerUrl` set to the configured start URL and goes straight to `oidc.<region>.amazonaws.com/authorize`. Confirmed working on 0.39.0.

Worth flagging that #923 pitched PKCE purely as a UX improvement and deliberately kept device code as the default. It's also currently the only working path for dual-stack instances, which is maybe an argument for revisiting that default, or at least documenting it.

## Suggested fix

Load shared config instead of a bare one, so the standard AWS env vars and `~/.aws/config` settings apply:

```go
cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(rootProfile.SSORegion()))
```

Same pattern applies at the other `aws.NewConfig()` sites (`pkg/cfaws/creds.go:111`, `pkg/cfaws/assumer_aws_sso.go:227`, `pkg/granted/doctor/doctor.go:228`) if AWS config is meant to be honored there too.

granted 0.39.0 (homebrew), macOS 26.5 arm64.

sbldevnet:
Fixes #960.

## Problem

`use_dualstack_endpoint` / `AWS_USE_DUALSTACK_ENDPOINT` (and the FIPS
equivalents) had no effect on any SSO call. Users on dual-stack Identity Center
instances kept receiving legacy IPv4-only portal URLs with no error or warning.

Every SSO client was built from a bare `aws.NewConfig()`, which leaves
`ConfigSources` empty. The generated resolvers bail out on exactly that
condition — `ssooidc@v1.35.15/api_client.go:729`:

```go
func resolveUseDualStackEndpoint(cfg aws.Config, o *Options) error {
	if len(cfg.ConfigSources) == 0 {
		return nil
	}
	...
}
```

`ConfigSources` records where configuration was read from, and only
`config.LoadDefaultConfig` populates it. So the setting was parsed from
`~/.aws/config`, then dropped before it could reach the client.

`resolveUseFIPSEndpoint` (line 744) has the identical guard, which is why FIPS
was silently ignored too.

The portal domain is chosen by the OIDC endpoint, so this is what the user saw —
same start URL, both endpoints probed directly:

```
oidc.eu-west-1.amazonaws.com  -> https://ssoins-....eu-west-1.portal.amazonaws.com/#/device?...
oidc.eu-west-1.api.aws        -> https://ssoins-....portal.eu-west-1.app.aws/#/device?...
```

## Fix

Adds `cfaws.LoadSSOConfig` and uses it in place of `aws.NewConfig()` at the four
sites that build SSO clients:

| Site | Client |
|------|--------|
| `pkg/cfaws/assumer_aws_sso.go` | `ssooidc` (login), `sso` (`GetRoleCredentials`) |
| `pkg/granted/sso.go` | `ssooidc` (`granted sso login`) |
| `pkg/granted/doctor/doctor.go` | `sso` (`granted doctor`) |

`WithSharedConfigProfile` is passed where a profile is known — without it the
SDK reads the setting from the *default* profile rather than the one being
assumed. `granted sso login` has no profile, so environment variables and the
default profile are its only sources.

Loading is best effort: an unreadable or missing profile falls back to the
previous region-only behaviour rather than failing the login.

`pkg/cfaws/creds.go` is deliberately untouched. It builds an `sts` client with
explicitly injected static credentials to validate them; a shared-config profile
would be wrong there, and it has no bearing on portal URLs.

## Behaviour changes worth reviewing

`LoadDefaultConfig` brings in the whole shared configuration, not just
dual-stack, so settings that were previously ignored are now honoured:

- **FIPS.** The `ssooidc` ruleset emits `oidc-fips.<region>.amazonaws.com`
  whenever the partition supports FIPS, and the `aws` partition declares that
  globally. A user with `use_fips_endpoint = true` set globally in a region with
  no OIDC FIPS endpoint currently logs in fine, and after this change would fail
  DNS resolution. This is the setting being obeyed as documented, but it is a
  visible change.
- **`endpoint_url` / `AWS_ENDPOINT_URL`** now applies to SSO calls. Combined
  with dual-stack or FIPS the ruleset rejects it outright:
  `Invalid Configuration: Dualstack and custom endpoint are not supported`.
- **Empty SSO region.** `WithRegion("")` no longer pins an empty region, so the
  SDK falls through to `AWS_REGION` or the profile's `region`.

Happy to gate any of these differently if maintainers prefer.

## Not included

Auto-enabling dual-stack for `.app.aws` start URLs. A user who sets nothing
still gets the legacy portal URL. That would mean granted overriding resolved
AWS configuration from a heuristic on a URL string, inserting a rule into the
SDK's precedence chain that no other AWS tool has — `aws sso login` with the
same profile would still use the legacy endpoint. It seems better left to AWS,
which issues the `.app.aws` domain in the first place. A warning when the start
URL looks dual-stack but the setting is unset would be a smaller alternative if
there is appetite for it.

## Testing

`pkg/cfaws/ssoconfig_test.go` asserts the state that actually reaches the
client, `ssooidc.NewFromConfig(cfg).Options().EndpointOptions.UseDualStackEndpoint`,
against a temporary `AWS_CONFIG_FILE`: profile opt-in, environment opt-in,
profile without opt-in, unknown profile falling back to the environment, and
nothing set. The three opt-in cases fail against the previous behaviour.

Verified manually against a real dual-stack Identity Center instance:
`granted sso login` now returns an `.app.aws` `verificationUriComplete` with
`AWS_USE_DUALSTACK_ENDPOINT=true`, and the legacy URL without it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/pkg/cfaws/ssoconfig.go
+++ b/pkg/cfaws/ssoconfig.go
@@ -11,11 +11,13 @@ import (
 // LoadSSOConfig loads the shared AWS configuration for the sso and ssooidc
 // clients, which ignore settings such as use_dualstack_endpoint unless the
 // config records the sources it was read from. profile may be empty, and
-// loading is best effort.
+// loading is best effort. BaseEndpoint is cleared so a custom endpoint_url
+// set for other AWS calls can't redirect SSO traffic.
 func LoadSSOConfig(ctx context.Context, region string, profile string) aws.Config {
 	if profile != "" {
 		cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(region), config.WithSharedConfigProfile(profile))
 		if err == nil {
+			cfg.BaseEndpoint = nil
 			return cfg
 		}
 		clio.Debugw("could not load shared config for profile, falling back to the default profile", "profile", profile, "error", err)
@@ -26,5 +28,6 @@ func LoadSSOConfig(ctx context.Context, region string, profile string) aws.Confi
 		clio.Debugw("could not load shared config, endpoint settings will be ignored", "error", err)
 		return aws.Config{Region: region}
 	}
+	cfg.BaseEndpoint = nil
 	return cfg
 }
--- a/pkg/cfaws/ssoconfig.go
+++ b/pkg/cfaws/ssoconfig.go
@@ -1,0 +1,30 @@
+package cfaws
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/config"
+	"github.com/common-fate/clio"
+)
+
+// LoadSSOConfig loads the shared AWS configuration for the sso and ssooidc
+// clients, which ignore settings such as use_dualstack_endpoint unless the
+// config records the sources it was read from. profile may be empty, and
+// loading is best effort.
+func LoadSSOConfig(ctx context.Context, region string, profile string) aws.Config {
+	if profile != "" {
+		cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(region), config.WithSharedConfigProfile(profile))
+		if err == nil {
+			return cfg
+		}
+		clio.Debugw("could not load shared config for profile, falling back to the default profile", "profile", profile, "error", err)
+	}
+
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(region))
+	if err != nil {
+		clio.Debugw("could not load shared config, endpoint settings will be ignored", "error", err)
+		return aws.Config{Region: region}
+	}
+	return cfg
+}
--- a/pkg/cfaws/assumer_aws_sso.go
+++ b/pkg/cfaws/assumer_aws_sso.go
@@ -190,8 +190,7 @@ func (c *Profile) SSOLogin(ctx context.Context, configOpts ConfigOpts) (aws.Cred
 	}
 
 	if cachedToken == nil && plainTextToken == nil {
-		newCfg := aws.NewConfig()
-		newCfg.Region = rootProfile.SSORegion()
+		newCfg := LoadSSOConfig(ctx, rootProfile.SSORegion(), rootProfile.Name)
 
 		var newSSOToken *securestorage.SSOToken
 		var err error
@@ -203,9 +202,9 @@ func (c *Profile) SSOLogin(ctx context.Context, configOpts ConfigOpts) (aws.Cred
 			clio.Warn("Headless environment detected, falling back to device code flow instead of authorization code")
 		}
 		if configOpts.UseAuthorizationCode && !useDeviceCode {
-			newSSOToken, err = idclogin.LoginWithAuthorizationCode(ctx, *newCfg, rootProfile.SSOStartURL(), rootProfile.SSOScopes(), configOpts.SSOBrowserProfile)
+			newSSOToken, err = idclogin.LoginWithAuthorizationCode(ctx, newCfg, rootProfile.SSOStartURL(), rootProfile.SSOScopes(), configOpts.SSOBrowserProfile)
 		} else {
-			newSSOToken, err = idclogin.Login(ctx, *newCfg, rootProfile.SSOStartURL(), rootProfile.SSOScopes(), configOpts.SSOBrowserProfile)
+			newSSOToken, err = idclogin.Login(ctx, newCfg, rootProfile.SSOStartURL(), rootProfile.SSOScopes(), configOpts.SSOBrowserProfile)
 		}
 		if err != nil {
 			return aws.Credentials{}, err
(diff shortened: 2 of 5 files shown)


pkg/cfaws/ssoconfig.go
```diff
@@ -0,0 +1,30 @@
...
+		}
+		clio.Debugw("could not load shared config for profile, falling back to the default profile", "profile", profile, "error", err)
+	}
+
+	cfg, err := config.LoadDefaultConfig(ctx, config.WithRegion(region))
+	if err != nil {
+		clio.Debugw("could not load shared config, endpoint settings will be ignored", "error", err)
+		return aws.Config{Region: region}
+	}
+	return cfg
```
meyerjrr, line 32: If the `AWS_ENDPOINT_URL` and region are going to be problems here, it might be reasonable to override their values to keep the existing user experience?

```
cfg.BaseEndpoint = nil
cfg.Region = region
```

Region would be less of an issue but unsetting the base endpoint seems like a requirement here wdyt?

sbldevnet: I did some tests and the AWS CLI turns out to do the same thing granted does with this PR. With `AWS_ENDPOINT_URL` set, it sends the SSO login there too.

So someone with a global `AWS_ENDPOINT_URL` set for something else already has a broken `aws sso login` today, while granted kept working because it ignored the setting entirely. This PR makes granted stop being the exception. And AWS already ships the opt-out for exactly that situation (`AWS_IGNORE_CONFIGURED_ENDPOINT_URLS=true`, or `ignore_configured_endpoint_urls = true` in the profile), which I verified works in both tools.

JoshuaWilkes (approved):

pkg/cfaws/ssoconfig.go
meyerjrr: Lets leave granted in the same state it was in previously by unsetting the `BaseEndpoint`. We can fix forward with a config field like `AWS_IGNORE_CONFIGURED_ENDPOINT_URLS` if issues crop up and some users want it set

meyerjrr (approved):

PR #973, opened 2026-09-06
 Migrate prompts from survey to huh

meyerjrr:
The library we used for multiple choice selections:`AlecAivazis/survey` was archived since 2023. This PR replaces it with huh v2. 
Builds on #956 by @nf-matt (RFC in #945) different API shape per the review there.

Call sites use huh directly. But all use a consistent `Form` which enforces stderr output + theme.

Also drops `pkg/testable` an unused testing surface.

### Behaviour changes
New library brings some new behaviour changes, I have intentionally not changes any of the default functionality to minimise the amount of custom logic we need to make the migration off survey.

- Filtering is `/` then type, plus vim motions (huh default). Was: any keypress filtered.
- Multi-token filter gone — `prod eu` no longer matches `eu-west-prod`. huh's filter is
  substring-only and not pluggable.
- Ctrl+C cancels, ESC unbound.
- Menus cap at 12 lines and scroll, so warnings printed above stay on screen.
- `settings set` on an int field now works, was broken before (string into `reflect` int).

Recording 2026-09-06 at 17 51 12


pkg/assume/assume.go
```diff
@@ -590,93 +580,63 @@ func printFlagUsage(region, service string) {
...
 	for _, pn := range profileNames {
 		if len(pn) > longestProfileNameLength {
 			longestProfileNameLength = len(pn)
 		}
 	}
+
+	nameColumn := "%-" + strconv.Itoa(longestProfileNameLength) + "s%s"
 	lightBlack := ansi.ColorFunc(ansi.LightBlack)
 	var hasDescriptions bool
+	options := make([]huh.Option[*cfaws.Profile], len(profileNames))
```
meyerjrr, line 611: side effect benefit here that we can pass the whole profile as the options array. no need to map the selected profile string name to the profile object separately

